PR #26, opened 2020-07-17
finished mongodb example

elianiva:
I've finished the example for mongodb. To set this up, all you have to do is add `DB_URI` on `.env` file and you're good to go.
It has 4 methods which are `GET`, `POST`, `DELETE`, and `PUT`. It uses `req.query` instead of `req.body` so make sure to not get it wrong.
`GET /notes` will return a list of notes with 2 properties which are title and desc.
`POST /notes` requires you to enter 2 query which are title and desc.  
`DELETE /notes` only need an ID of that note.
`PUT /notes` requires and ID of the item that you want to update, as well as title and desc for the updated one.

### DIFF
--- a/examples/mongodb/index.js
+++ b/examples/mongodb/index.js
@@ -1,12 +1,71 @@
 import { App } from '@tinyhttp/app'
-import { json } from 'body-parsec'
+import dotenv from 'dotenv'
+import mongodb from 'mongodb'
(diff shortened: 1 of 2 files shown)


talentlessguy:
LTGM, merging this

talentlessguy:
@all-contributors please add  elianiva for example

PR #29, opened 2020-07-18
refactor: update dev-server example

katywings:
Updating the dev-server example so that it:
- Reloads the server on changes
- Reloads the client
- Shows server errors on the client

P.S. i am not entirely sure whats the advertised use case with this example, but `serve-handler` seems to be a very "strict" static server middleware - it directly outputs a 404 if it fails to find a file for any request - instead of giving other middlewares the chance to do something better :D, so this makes it imho hard to use along with tinyhttp.


talentlessguy:
LGTM, merging

talentlessguy:
@all-contributors please add katywings for example

PR #32, opened 2020-07-18
Add serverless example + lower engines

talentlessguy:
Add Serveless example and lower the engine in packages to work on Vercel


talentlessguy:
@all-contributors please add talentlessguy for example, plugin, projectManagement, example, maintenance

PR #40, opened 2020-07-19
Output configuration for logger

### LINKED ISSUE #38: Use a logging library as a logger implementation instead of console.log
**Is your feature request related to a problem? Please describe.**
I feel that using something like `pino` or `winston` will be best for logging purposes, since there is no need to create our own logger and reinvent the wheel.

**Describe the solution you'd like**
Rewrite the current logger package to use a fully-featured logging library.

**Describe alternatives you've considered**
Using `console.log` has disadvantages, some users would like to save the logs to a file, log via http, etc... This is not possible using our current logging solution.

Arnovsky:
closes #38 

* option to provide output function.
* option to chose if to color the message (useful when logging to a file)
* changed `||` to `??` in the logger package


packages/logger/README.md
```diff
@@ -24,6 +24,8 @@ Returns the middleware to log HTTP requests.
 
 - `methods` - a list of HTTP methods to log. Defaults to `http`'s `METHODS`
 - `timestamp.format` - timestamp format. Gets consumed by [dayjs](https://day.js.org) library. If a string specified, used as a format, otherwise just eables it.
+- `outputConfiguartion.output` - function that recieved the log generated by the logger package
+- `outputConfiguartion.color` - property that determines if the logger will generate a message with color. useful for logging into the console, disable it if logging into file / http / etc...
```
talentlessguy, line 28: you have a typo in README

packages/logger/README.md
```diff
@@ -34,7 +36,7 @@ import { logger } from '@tinyhttp/logger'
 const app = new App()
 
 app
-  .use(logger({ methods: ['GET', 'POST'], timestamp: { format: 'HH:mm:ss' } }))
+  .use(logger({ methods: ['GET', 'POST'], timestamp: { format: 'HH:mm:ss' }, outputConfiguration: { output: console.log, color: false } }))
```
talentlessguy, line 39: What about changing `outputConfiguration` to `output`, and `output` to `function`?

packages/logger/src/index.ts
```diff
@@ -4,16 +4,21 @@ import { IncomingMessage as Request, ServerResponse as Response, METHODS } from
 
 export type LoggerProperties = Partial<{
   methods: string[]
+  outputConfiguration: {
```
talentlessguy, line 7: Every property in the config must be optional, here `color` and `output` seem to be both required.

EIther put `?` in the end of properties names or use `DeepPartial` ([ref](https://stackoverflow.com/a/49936686))

packages/logger/README.md
Arnovsky: I thought `callback` might me better, but I think both are ambiguous.

talentlessguy: @Arnovsky hmm `callback` sounds good

packages/logger/src/index.ts
Arnovsky: Wouldn't this mean changing `LoggerProperties` to an interface? since there is no meaning when using `Partial` combined with `?`

talentlessguy: @Arnovsky hmm then probably u're right, it makes more sense to change it to interface.

Also, please cover ur new functionality in tests (if there are none yet) :)

packages/logger/src/index.ts
```diff
@@ -2,18 +2,19 @@ import colors from 'colors'
...
 
-export type LoggerProperties = Partial<{
-  methods: string[]
-  timestamp:
-    | {
-        format?: string
-      }
-    | boolean
-}>
+interface LoggerProperties {
```
talentlessguy, line 5: Could you please export the interface, and rename it to `LoggerOptions`?

talentlessguy:
@all-contributors please add Arnovsky for code and plugin

Issue #24, opened 2020-07-17
Add the rest of HTTP methods to Router


talentlessguy:
**Is your feature request related to a problem? Please describe.**

At the moment [Router](https://github.com/talentlessguy/tinyhttp/blob/master/packages/app/src/router.ts) only has a few methods that can be used. There must be all of them.

**Describe the solution you'd like**

Copy paste some method and change it's name.

to get a full list of methods:

```sh
node -p 'http.METHODS'
```

**Additional context**

[router.ts](https://github.com/talentlessguy/tinyhttp/blob/master/packages/app/src/router.ts) file

talentlessguy:
Completed with #39 and #44 :tada:

PR #44, opened 2020-07-20
added some route tests

elianiva:
I've added some http method test in continuation of #24. Some of them are `GET`, `POST`, `PUT`, `DELETE`, etc.


talentlessguy:
@all-contributors add elianiva for test

PR #57, opened 2020-08-28
Make the pages responsive

RocktimSaikia:
Made the pages responsive especially the code sample sections that were causing inconsistency in the responsiveness. The overall design structure is not that great. Many unnecessary width settings inside the same parent element when one width setting in the parent can fix most of the responsive issues. For now, it should do but later will take a good look on that I guess.


talentlessguy:
Can you send a screenshot of how it looks like on mobile please?

RocktimSaikia:
1. These code snippets were overflowing before. Now they don't.
2. Also fixed some minor width issues along with the `localhost browser view` which looks much cleaner now.




 [ps: I see many things that can be improved on the site from colors, contrasts to font sizes and all But maybe later. Currently, the focus is on the actual framework if I'm not wrong ] @talentlessguy

site/static/css/docs.css
```diff
@@ -14,13 +14,14 @@ pre {
   overflow-x: auto;
   box-shadow: none;
   padding: 1rem;
+  border-radius: 5px;
 }
 
-main > h1 {
+main>h1 {
```
talentlessguy, line 20: Your Prettier in VSCode doesn't format CSS properly, please fix this

site/static/css/docs.css
```diff
@@ -5,7 +5,7 @@ main {
 
 @media (max-width: 350px) {
   main {
-    width: 80vw;
+    width: 90vw;
```
talentlessguy, line 8: Not sure why we need to make it wider?

talentlessguy (changes requested):

site/static/css/docs.css
RocktimSaikia: I think the whitespaces around was taking too much space. Kinda squeezing the code snippet
.

This much width makes it look much more readable imo . Don't you think ?

site/static/css/docs.css
RocktimSaikia: I don't have prettier. I only use the `XO` plugin for now. I was contributing to one of [sindresorhus](https://github.com/sindresorhus) projects where my prettier plugin was causing some errors so had to remove it and use xo plugin. I'll have to switch back I guess then

RocktimSaikia: @talentlessguy fixed!

site/static/css/docs.css
talentlessguy: hm yeah this makes sense

site/static/css/docs.css
talentlessguy: @RocktimSaikia your editor must be using a prettier and eslint config defined in the project root (for the future), but it looks good so merging

talentlessguy:
@all-contributors please add RocktimSaikia for code

PR #64, opened 2020-08-31
:sparkles: implement req.ip and req.ips

RocktimSaikia:
implement `req.ip` and `req.ips` . The implementation is a bit different from `express`.
- There was already an address `trust` function implemented in our codebase. and Both of these methods rely on that for making sure that the incoming address is trustworthy.
- `req.ips` returns all the `IP addresses` when the request is coming under a proxy. And when it there is no proxy it just returns the first [untrusted IP](https://www.npmjs.com/package/proxy-addr#proxyaddrallreq-trust) which is `127.0.0.1` in the local server
- `req.ip` method has a regex replacer in its return statement which removes the prefix added by the [OS to `IPv4` to convert it to `IPv6`](https://stackoverflow.com/a/35560645/9360234). Basically it just prefixes `:ffff:` in front of the IPv4. So that regex replacer removes this part from IP address

***test cases for both of these methods are added***

### DIFF
--- a/__tests__/request.test.ts
+++ b/__tests__/request.test.ts
@@ -23,6 +23,34 @@ describe('Request properties', () => {
       done
     )
   })
+
+  it('req.ip & req.ips is being parsed properly', (done) => {
+    const { request } = InitAppAndTest(
+      (req, res) => {
+        res.json({
+          ip: req.ip,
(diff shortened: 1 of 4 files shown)


talentlessguy:
epic, merging

PR #82, opened 2020-09-25
Update index.html

tkalmi:
I assume that Express middleware works with tinyhttp, which is what the altered line is trying to say.


talentlessguy:
ah yes, my bad, and thanks for the fix

talentlessguy:
@all-contributors please add @tkalmi for doc (I guess?)

PR #97, opened 2020-10-01
Unit testing: uncommented accepts tests and added another one

BRA1L0R:
### Added:
- types unit test for checking that it returns the first option when there's no header

### Fixes:
- uncommented accept.types() tests


talentlessguy:
LGTM

Issue #93, opened 2020-10-01
Add Svelte (or Sapper) SSR example


talentlessguy:
Svelte SSR Example (w/ pure `http`): https://github.com/Rich-Harris/svelte-ssr-bundle

Sapper can also be used, here are the docs for it: https://sapper.svelte.dev/docs

To get started with contributing an example, check the [CONTRIBUTING.md](https://github.com/talentlessguy/tinyhttp/blob/master/CONTRIBUTING.md#adding-new-example) file

Deadik:
I'm interested, could you assign me?

talentlessguy:
@Deadik yep, sure :)

talentlessguy:
Closed via #135

PR #100, opened 2020-10-01
#93: Add svelte ssr example

Deadik:
Add simple example with svelte ssr

### DIFF
--- a/examples/svelte-ssr/index.js
+++ b/examples/svelte-ssr/index.js
@@ -1,0 +1,23 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+require('svelte/register')
+const fs = require('fs')
+const path = require('path')
+const { App } = require('@tinyhttp/app')
+const AppSvelte = require('./App.svelte').default
+
+server
+  .get('*', async (req, res) => {
+    const { url } = req
+    const { html, head } = AppSvelte.render({
+      url,
+      title: 'Svelte SSR',
+      htmlString: `
+            <h2>This is html string</h2>
+        `,
+    })
+
+    const indexFileContent = fs.readFileSync(path.resolve(__dirname, 'index.html'))
+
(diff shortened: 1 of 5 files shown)


examples/svelte-ssr/index.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+require('svelte/register')
```
talentlessguy, line 2: examples must use `import` / `export` syntax

examples/svelte-ssr/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "tinyhttp-svelte-ssr",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
```
talentlessguy, line 5: you forgot to put `"type": "module"`

talentlessguy (changes requested):

examples/svelte-ssr/index.js
Deadik: Svelte has some problems with export esm `import 'svelte/register'` or `import 'svelte/compiler'` throw "module not found".
Haven’t come up with a solution yet.

talentlessguy: gonna try it locally and see if it works

talentlessguy: it works if you specify a file extension, e.g.:

```js
import 'svelte/register.js'
```

Deadik:
#decline

talentlessguy:
@Deadik did something happen? you closed the pr for some reason :c

Deadik:
I'm not found working solution and can't spend more time

talentlessguy:
@Deadik specifying the file extension worked

but if you have no time, thanks for your effort :D 

someone else might get it working

Deadik:
specifying the file extension worked, but other my code with import not worked, should be changed to directly using svelte API for compile and render. A don't find how do this

PR #101, opened 2020-10-01
Add prisma example

Luiginator:
This PR adds an example for Prisma as requested [here](https://github.com/talentlessguy/tinyhttp/issues/95)


examples/prisma/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "private": true,
+  "name": "tinyhttp-prisma",
+  "type": "module",
+  "scripts": {
+    "start": "node -r esm -r ts-node/register ./src/index.ts"
+  },
+  "dependencies": {
+    "@tinyhttp/app": "workspace:*",
```
Luiginator, line 9: Should we use a published version instead? This makes sure that people that use `npm`or `yarn` can also run the example. E.g. the `vercel` example points to a released version.

talentlessguy: we shouldn't because it's a pnpm monorepo

when someone wants to setup an example he/she just needs to remove the `workspace` thing

In the future there will be a CLI to setup an example with a npm version by default

Luiginator: @talentlessguy ok then the PR is ready for your review.

examples/prisma/README.md
```diff
@@ -0,0 +1,47 @@
+# Prisma example
+
+A simple REST API using tinyhttp and Prisma
```
talentlessguy, line 3: Could you please wrap "Prisma" in a link to prisma site?

talentlessguy (changes requested):

talentlessguy:
Also, just a thought. Should the example use TypeScript? All of the rest examples are in plain JavaScript to show that they work normally without any extra compiling process

although, for Prisma I think it would be fine to have TypeScript used, so prob nothing should be changed, it's good already

Luiginator:
@talentlessguy good point. I removed TypeScript and implemented the requested change.

talentlessguy (approved):

talentlessguy:
alrigghhttt merging

talentlessguy:
@all-contributors please add @Luiginator for example

PR #118, opened 2020-10-03
Unless middleware

shzmr:
Unless middleware inspired by [express-unless](https://www.npmjs.com/package/express-unless) middleware.

### `unless(middleware, (UnlessMiddlewareOptions | CustomUnless))`

The `UnlessMiddlewareOptions` object can include:
- `method` - string or array of strings that describe forbidden http methods such as GET, POST, PUT etc...
- `path` - array of strings, Regex and objects that include `url` and `methods` properties, which will be compared against the request.
- `ext` - string or array of strings that describe forbidden path ends (e.g. in `/user/123` it will check against `/123`).

The `CustomUnless` is a function that receives a Request object and returns a boolean. The result of the function will determine if the execution of the middleware is skipped.


packages/unless/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+app.listen(3000);
+```
+
+## License
+
+MIT © [v1rtl](https://v1rtl.site)
+
+[npm-badge]: https://img.shields.io/npm/v/@tinyhttp/req?style=flat-square
```
talentlessguy, line 54: you should change the badge from `req` to `unless`

packages/unless/README.md
```diff
@@ -0,0 +1,56 @@
...
+//Path example
+app.use(unless(cors(), {path: ['/content/public', /user/, {url: "/public", methods: ['GET']}]});
+
+
+app.listen(3000);
+```
+
+## License
+
+MIT © [v1rtl](https://v1rtl.site)
```
talentlessguy, line 52: You wrote the module so I guess there should be a license for your name :)

packages/unless/package.json
```diff
@@ -0,0 +1,39 @@
...
+    },
+    "./package.json": "./package.json",
+    "./": "./"
+  },
+  "keywords": [
+    "tinyhttp",
+    "node.js",
+    "web framework",
+    "web",
+    "middleware"
```
talentlessguy, line 28: please add the `unless` tag :D

packages/unless/src/index.ts
```diff
@@ -0,0 +1,89 @@
+import type { Request, Response, NextFunction } from '@tinyhttp/app'
+
+//Options for the middleware
+export interface UnlessMiddlewareOptions {
```
talentlessguy, line 4: Not sure why it is an interface. It only contains properties, doesn't extend anything and nothing implements it. I'd use `type` instead, wrapped in the `Partial<T>`

talentlessguy (changes requested):

packages/unless/src/index.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+//Optional path object
+interface PathObject {
+  url: string
+  methods: string[]
+}
+
+//Middleware interface
+type Middleware = (req: Request, res: Response, next?: NextFunction) => any
```
talentlessguy, line 17: I forgot to re-export `Middleware` type in `app` package my bad

now it's re-exported so u just need to do `import type { Middleware } from '@tinyhttp/app'`

talentlessguy (changes requested):

shzmr:
Thanks, I will fix it.

talentlessguy:
LGTM, merging! 

will publish a bit later

talentlessguy:
@all-contributors add @shzmr for code

PR #120, opened 2020-10-03
Add tinyhttp + react example

### LINKED ISSUE #109: Create React SSR Example
Create a React SSR example with hydration on the client (e.g. react both on server and client) 

if it is too diffficult to do, then just SSR but PRs with client hydration will be in preference

You can use any abstract layers around SSR, react-ssr for example

React SSR docs: https://reactjs.org/docs/react-dom-server.html

`react-ssr` package: https://www.npmjs.com/package/react-ssr

setheal:
Resolve #109 
I don't used `react-ssr` and tried to create an example similar to the other ones present in the repository. Let me know if you need something different :)


talentlessguy:
Looks very nice! Even with client rehydration, amazing

merging!

talentlessguy:
@all-contributors add @setheal for example

PR #135, opened 2020-10-07
Added sapper example

shzmr:
Added svelte-sapper example, using the official template. Issue #93 .


examples/sapper/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "build": "sapper build --legacy",
+    "export": "sapper export --legacy",
+    "start": "node __sapper__/build"
+  },
+  "private": true,
+  "dependencies": {
+    "@tinyhttp/app": "workspace:*"
+  },
+  "devDependencies": {
+    "sapper": "^0.28.0",
```
talentlessguy, line 14: hm, why is sapper a dev dependency here, as well as svelte? should it be like this?

shzmr: This is what they do in the template. I think its because it is only used for building, and not for the program itself.

talentlessguy: @shzmr understood

talentlessguy:
@all-contributors please add @shzmr for example

Issue #140, opened 2020-10-08
Write tests for @tinyhttp/req


talentlessguy:
- [ ] `req.range`
  - [ ]  should return parsed ranges
  - [ ] should cap to the given size
  - [ ] should cap to the given size when open-ended
  - [ ] should have a .type
  - [ ] should accept any type
  - [ ] should return undefined if no range
  - [ ] with `options`
    - [ ] Should return combined ranges if `combine` set to true
- [ ] `req.is`
  - [ ] should return the given MIME type when matching
  - [ ] should return false when not matching
  - [ ] should return false when `Content-Type` header is not present
  - [ ] Should lookup the MIME type with the extension given (e.g. `req.is('json')`
  - [ ] Should ignore charset
  - [ ] more coming soon...
- [ ] more coming soon...

MaurizioPz:
Thanks

talentlessguy:
Closed via #142

atif-saddique-deel:
this also should have been covered in tests here.
https://github.com/tinyhttp/tinyhttp/issues/407

`req.route` always returns `undefined`
https://tinyhttp.v1rtl.site/docs#reqroute

talentlessguy:
> this also should have been covered in tests here.
> https://github.com/tinyhttp/tinyhttp/issues/407
> 
> `req.route` always returns `undefined`
> https://tinyhttp.v1rtl.site/docs#reqroute

you need to set `enableReqRoute` to true

PR #142, opened 2020-10-08
#140 added tests for req

MaurizioPz:
This PR adds the test listed on #140 

but the  test 'Should ignore charset' is ignored because right now it doesn't pass. I may have misinterpreted the desired test, but as of right now https://github.com/jshttp/media-typer seems to be using a regex that doesn't handle charset


talentlessguy:
@MaurizioPz by "should ignore charset" for `req.is` I meant that if you send a `Content-Type` that includes a charset, it must ignore it and check for the MIME type only

MaurizioPz:
Well than there's a bug right now, because it sounds like you are expecting this test to pass and it doesn't

```
    it('should ignore charset', async () => {
      const app = runServer((req, res) => {
        expect(reqIs(req)('text/html')).toBe('text/html')
        res.end()
      })
      await makeFetch(app)('/', {
        headers: {
          'Content-Type': 'text/html; charset=UTF-8',
        },
      })
      expect.assertions(1)
    })
```

BRA1L0R:
@MaurizioPz i've investigated on the issue and pointed it out in #150. You can work on it if you want as it is very easy to fix but also relatively important.

talentlessguy:
@MaurizioPz could you pls fix the conflict so we can merge this?

talentlessguy:
nice, now it looks good

merging

talentlessguy:
@all-contributors add @MaurizioPz for test

PR #152, opened 2020-10-10
added markdown example

HunterJS-bit:
add basic example for rendering markdown


talentlessguy:
Why does it use a logger and why it doesn't use `gray-matter`

talentlessguy:
A week passed, and it still doesn't include a `gray-matter` extraction

so closing

feel free to ask to reopen

PR #153, opened 2020-10-11
reqIs does not correctly parse the Content-Type header #150

### LINKED ISSUE #150: reqIs does not correctly parse the Content-Type header
Thanks @MaurizioPz for finding this bug

**Describe the bug**
`reqIs` does not separate the content type from the charset
This is actually an issue in another package called `@tinyhttp/type-is`
```ts
function tryNormalizeType(value: string) {
  if (!value) return null

  try {
    return normalizeType(value)
  } catch (err) {
    return null
  }
}
```

as you can see here (line 12, @tinyhttp/type-is/index.ts) the function for type normalization is empty and does not remove the charset part from the content-type header

**To Reproduce**
( related issue #142 )
Running this test results in an error, since type-is does not properly separate the actual content from the charset info.
```ts
it('should ignore charset', async () => {
    const app = runServer((req, res) => {
      expect(reqIs(req)('text/html')).toBe('text/html')
      res.end()
    })
    await makeFetch(app)('/', {
      headers: {
        'Content-Type': 'text/html; charset=UTF-8',
      },
    })
    expect.assertions(1)
  })
```

**Expected behavior**
type-is.typeIs should separate the `text/html` part of the Content-Type header from `; charset=UTF-8`

**Versions**
* `node`: 14.13.0
* `@tinyhttp/app`: 0.4.3

MaurizioPz:
To fix #150 I've replaced media-typer with another module from the same library (jshttp) that is more appropriate to parse a content type header (as stated in media-typer readme https://github.com/jshttp/media-typer#media-typer)

### DIFF
--- a/__tests__/modules/req.test.ts
+++ b/__tests__/modules/req.test.ts
@@ -1,5 +1,5 @@
 import { makeFetch } from 'supertest-fetch'
-import { checkIfXMLHttpRequest, getAccepts, getFreshOrStale, getRequestHeader } from '../../packages/req/src'
+import { checkIfXMLHttpRequest, getAccepts, getFreshOrStale, getRequestHeader, reqIs } from '../../packages/req/src'
 import { runServer } from '../../test_helpers/runServer'
 
 describe('Request extensions', () => {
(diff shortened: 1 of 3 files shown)


talentlessguy:
Could you please use [es-content-type](https://www.npmjs.com/package/es-content-type) instead? it's the same module but with types ootb

talentlessguy:
LGTM

talentlessguy:
@all-contributors add @MaurizioPz for code

PR #160, opened 2020-10-14
Chore - Send tests

### LINKED ISSUE #143: Write tests for @tinyhttp/send
- [ ] send
      - [ ] `send`
        - [ ] remove headers for 301 and 204 statuses ([codecov](https://codecov.io/gh/talentlessguy/tinyhttp/src/master/packages/send/src/send.ts#L45))
        - [ ] set `Content-Type` to `application/octet-stream` if the header hasn't been set before
      
      - [ ] `sendFile`
        - [ ] throw if path is not absolute ([codecov](https://codecov.io/gh/talentlessguy/tinyhttp/src/master/packages/send/src/sendFile.ts#L37))
        - [ ] `Content-Type` header (based on filename) is set properly
        - [ ] `options.headers` sets all required headers

alcar:
Closes #143.

### DIFF
--- a/packages/send/src/sendFile.ts
+++ b/packages/send/src/sendFile.ts
@@ -49,9 +49,11 @@ export const sendFile = <Request extends I = I, Response extends S = S>(_: Reque
 
   const stream = createReadStream(root ? root + path : path, options)
 
-  stream.on('error', (err) => void cb(err))
+  if (cb !== undefined) {
+    stream.on('error', (err) => void cb(err))
 
-  stream.on('end', () => void cb())
+    stream.on('end', () => void cb())
+  }
 
   res.setHeader('Content-Type', contentType(extname(path)))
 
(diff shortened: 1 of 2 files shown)


packages/send/src/sendFile.ts
```diff
@@ -49,9 +49,11 @@ export const sendFile = <Request extends I = I, Response extends S = S>(_: Reque
 
   const stream = createReadStream(root ? root + path : path, options)
 
-  stream.on('error', (err) => void cb(err))
+  if (cb !== undefined) {
```
talentlessguy, line 52: if cb is null it can also error. Better do non-strict checking like `cb != null` so it won't execute for neither `undefined`, nor `null` nor `false`

talentlessguy (changes requested):

talentlessguy (approved):

talentlessguy:
@all-contributors add @alcar for test

Issue #92, opened 2020-10-01
Tests TODO


talentlessguy:
The final goal is of this issue is to achieve ~80-100% coverage

here's a todo list of things that require tests (based on the `__tests__`  folder):

> :warning: List is in progress, I'll be adding new test tasks over time, here are some that you can submit already

- [ ] core
   - [ ] app
     - [x] `app` - `app.set`, `app.enable`, `app.disable` ([codecov](https://codecov.io/gh/talentlessguy/tinyhttp/src/master/packages/app/src/app.ts#L96))
     - [x] `onError` - Test a case where an error doesn't have an HTTP status code, neither is a string ([codecov](https://codecov.io/gh/talentlessguy/tinyhttp/src/master/packages/app/src/onError.ts#L14))
   - [ ] request
   - [ ] response
- [ ] modules
   - [x] accepts
   - [x] cookie-signature
   - [x] cookie
   - [x] etag (#133)
   - [x] req (#140)
   - [x] res
      - [x] `res.format` - if `default` only is specified, call it as a function ([codecov]())
      - [x] `res.format` - if nothing is specified, throw 406 ([codecov](https://codecov.io/gh/talentlessguy/tinyhttp/src/master/packages/res/src/format.ts#L28))
   - [ ] router
   - [x] send (#143)
- [x] wares
   - [x] cookie-parser
   - [x] cors
     - [x] Accept a function to set origin
   - [x] dotenv
   - [x] favicon
   - [x] jwt
     - [x] Should work with different algorithms ([codecov](https://codecov.io/gh/talentlessguy/tinyhttp/src/master/packages/jwt/src/index.ts#L30)) [pending]
   - [x] logger
   - [x] ping
   - [x] bot-detector (#111) 
   - [x] ip-filter
   - [x] jsonp (#123)\

If you want to help with tests, write down to this thread with your questions

talentlessguy:
Thanks for improved coverage for `accepts` by @BRA1L0R via #97

aneeshrelan:
Can I try adding tests for bot-detector?

talentlessguy:
@aneeshrelan yup, will create a separate issue for that now

talentlessguy:
@aneeshrelan check https://github.com/talentlessguy/tinyhttp/issues/111

aneeshrelan:
@talentlessguy I see some tests for the mentioned wares, are they still needed?

shubhi23994:
I can take up some tests

[...]

PR #165, opened 2020-10-16
added test for pug package

KoljaTM:
Tests TODO #92
- added a test for the pug package


__tests__/wares/pug.test.pug
```diff
@@ -0,0 +1,6 @@
+doctype html
```
talentlessguy, line 1: for test files there is a folder called "fixtures"

talentlessguy (changes requested):

__tests__/wares/pug.test.ts
```diff
@@ -5,7 +5,9 @@ import { renderFile } from 'pug'
 
 describe('PUG templates tests', () => {
   it('should render pug template', async () => {
-    const { fetch, app } = InitAppAndTest((_req, res) => void res.render('../__tests__/wares/pug.test.pug', null, { renderOptions: { pageTitle: 'Test title', foo: 'bar' } }))
+    const { fetch, app } = InitAppAndTest(
+      (_req, res) => void res.render('../__tests__/fixtures/views/pug.test.pug', null, { renderOptions: { pageTitle: 'Test title', foo: 'bar' } })
```
talentlessguy, line 9: fixture files don't have to have `.test.` in their filename. It's for tests. for fixtures just use `test.pug` or smth like this

__tests__/wares/pug.test.ts
```diff
@@ -5,7 +5,9 @@ import { renderFile } from 'pug'
 
 describe('PUG templates tests', () => {
```
talentlessguy, line 6: Pug is a name so it should be capital case

talentlessguy (changes requested):

__tests__/wares/pug.test.ts
```diff
@@ -3,11 +3,9 @@ import { createServer } from 'http'
 import { InitAppAndTest } from '../../test_helpers/initAppAndTest'
 import { renderFile } from 'pug'
```
talentlessguy, line 4: why do you import `renderFile` here?

KoljaTM: i had been trying out something and forgot to remove the imports. Is there a specific reason to set
   "noUnusedLocals": false
in the tsconfig?

talentlessguy: i think i accidentally copied it from my another project, gonna remove the setting

PR #182, opened 2020-11-03
lazy isbot

omrilotan:
Make isBot and botName lazy and idempotent:

- Will only calculate when requested
- Will only calculate once per request


talentlessguy:
LGTM, merging

talentlessguy:
@all-contributors add @omrilotan for code

PR #188, opened 2020-11-17
OpenAPI example

### LINKED ISSUE #186: Create OpenAPI validator example
Create an example using [express-openapi-validator](https://github.com/cdimascio/express-openapi-validator) and tinyhttp

MVEMCJSUNPE:
Should hopefully resolve #186 

I decided to go with [express-openapi-validate](https://github.com/Hilzu/express-openapi-validate), since I didn't see any examples of middleware in [openapi-validator](https://github.com/kogosoftwarellc/open-api/tree/master/packages/openapi-response-validator)


examples/express-openapi-validator/README.md
```diff
@@ -0,0 +1,20 @@
+# OpenAPI validator example
+
+Example of how to use express-openapi-validate with tinyhttp
```
talentlessguy, line 3: could you please wrap `express-openapi-validate` in a link to the package (I mean it's github repo?)

thanks in advance

examples/express-openapi-validator/app.js
```diff
@@ -0,0 +1,77 @@
+import { App } from '@tinyhttp/app'
+
+import fs from 'fs'
+import { OpenApiValidator } from 'express-openapi-validate'
+import jsYaml from 'js-yaml'
+import bodyParser from 'body-parser'
```
talentlessguy, line 6: not that much required but could you please use [milliparsec](https://github.com/talentlessguy/milliparsec) instead of body-parser?

packages/markdown/src/index.ts
```diff
@@ -47,7 +47,7 @@ export const markdownStaticHandler = (
 
         const isDir = !(dir === '')
 
-        return /\.(md|markdown)/.test(ext) && unPrefixedURL === (isDir ? `${dir}/${name}` : name)
+        return /\.(md|markdown)/.test(ext) && unPrefixedURL === (isDir ? `${dir}/${name}`.replace('\\', '/') : name)
```
talentlessguy, line 50: why did you add this and how it's related to OpenAPI validation?

talentlessguy (changes requested):

packages/markdown/src/index.ts
MVEMCJSUNPE: It's not related to OpenAPI validation. I'm using Windows, and I can't commit unless all the tests pass due to some pre-commit hooks. Markdown uses recursiveReadDir or promise from readdirp, which will print paths differently based on the operating system. Windows uses backslashes in its directory path, while Unix uses slashes. For example, on Windows:
```
PS C:\Users\sisyphus\source\repos\tinyhttp\packages\markdown\src> node
Welcome to Node.js v15.2.0
Type ".help" for more information.
var recursiveReadDir = require('readdirp').promise;
undefined
> recursiveReadDir('.').then((value) => { console.log(value); })
Promise { <pending> }
> [
  {
    path: 'index.ts',
    fullPath: 'C:\\Users\\sisyphus\\source\\repos\\tinyhttp\\packages\\markdown\\src\\index.ts',
    basename: 'index.ts',
    dirent: Dirent { name: 'index.ts', [Symbol(type)]: 1 }
  }
]
```
If we look at the following segment:
https://github.com/talentlessguy/tinyhttp/blob/b78a2d60766b4d22c9e55053c09d1d95d942ce45/packages/markdown/src/index.ts#L45-L50

we see that unprefixedUrl is being compared with the expression ``(isDir ? `${dir}/${name}` : name)``, but unprefixedUrl is a parsed url (`/talentlessguy/tinyhttp/pull/188` for example), which always uses forward slashes. So if we use the URL of this github pull request for example, the code would be comparing `/talentlessguy/tinyhttp/pull/188` with `\\talentlessguy\\tinyhttp\\pull\\188`; on Windows this results in an error for one of the markdown tests.

MVEMCJSUNPE: So the replace is needed to change backslashes into slashes so when testing on Windows, `unprefixedUrl` and `(isDir ? `${dir}/${name}` : name)` equal each other for a valid GET request accessing a markdown file that exists.

talentlessguy: ah, no problem then 

thanks for fixing the bug! :D

talentlessguy (approved):

talentlessguy:
@all-contributors add @MVEMCJSUNPE for example and for bug

PR #195, opened 2020-12-04
Remove undefined from NextFunction

bragle:
It seems like tinyhttp is incompatible with Express' NextFunction, making types for middleware that was originally designed for Express invalid. If this is intentional or if I'm wrong, feel free to just close this PR 😸

### DIFF
--- a/packages/router/src/index.ts
+++ b/packages/router/src/index.ts
@@ -2,7 +2,7 @@ import { METHODS, IncomingMessage as I, ServerResponse as R } from 'http'
 
 /* HELPER TYPES */
 
-export type NextFunction = (err?: any) => void | undefined
+export type NextFunction = (err?: any) => void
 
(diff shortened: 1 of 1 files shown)


talentlessguy:
could you please show an example of where it throws a type error because of that?

I put `undefined` there because some wares might not have `next`

bragle:
@talentlessguy sure thing!

Here's a few examples: https://codesandbox.io/s/strange-elgamal-q87is?file=/src/index.ts

Also, here's typings for Express' NextFunction (no `undefined`): https://github.com/DefinitelyTyped/DefinitelyTyped/blob/fe18c07ed84e3b0906b126892a1ae05ed351390b/types/express-serve-static-core/index.d.ts#L34

Hope this helps 😺

talentlessguy:
ok, I see, merging then

talentlessguy:
@all-contributors add @bragle for code

PR #225, opened 2021-02-04
Ability to send null .json responses

calumk:
this should fix the issue https://github.com/talentlessguy/tinyhttp/issues/223 where res.json() cant send null responses.

PR #226, opened 2021-02-04
Update send.test.ts

calumk:
First attempt at writing a test. likely to fail!


talentlessguy:
for some reason tests are hanging... and could you please instead of opening a new PR make a commit to your original (#225) PR?

Issue #229, opened 2021-02-09
CORS: 404 request for all OPTION requests


tim-field:
**Describe the bug**

I get a 404 for OPTIONS requests, that work with GET.

I've tried `preflightContinue: true` that didn't help either

**To Reproduce**

```
app
  .use(cors({origin: '*'}))
  .options('*', cors({origin: '*'}))
  .get('/collection', async (_, res) => {
 ```

**Versions**

* `node`: 14
* `@tinyhttp/cors`: 1.1.5
* `@tinyhttp/app`: 1.1.12

**Comment by talentlessguy** [comment]
@tim-field do you have an idea of how to fix this?

**Comment by aldy505** [comment]
@tim-field can you please try using #231 or through my sample repo here: https://github.com/aldy505/tinyhttp-cors-test

**Comment by talentlessguy** [comment]
@aldy505 should be fixed with the new version:

```sh
pnpm i @tinyhttp/cors@1.1.7
```

PR #231, opened 2021-02-11
fix: #229 - CORS response on preflight

aldy505:
fix!: preflightContinue defaults to false again now
fix: added preflight just for options method (as per express' cors)
chore: bump version

I've put a sample repo with this version on https://github.com/aldy505/tinyhttp-cors-test

### DIFF
--- a/packages/cors/src/index.ts
+++ b/packages/cors/src/index.ts
@@ -60,8 +60,7 @@ export const cors = (opts: AccessControlOptions = {}) => {
     // Setting the Access-Control-Max-Age header
     if (maxAge) res.setHeader('Access-Control-Max-Age', maxAge)
 
-    const method = req.method && req.method.toUpperCase && req.method.toUpperCase()
-    if (method === 'OPTIONS') {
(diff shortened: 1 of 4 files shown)


packages/cors/src/index.ts
```diff
@@ -60,11 +60,17 @@ export const cors = (opts: AccessControlOptions = {}) => {
     // Setting the Access-Control-Max-Age header
     if (maxAge) res.setHeader('Access-Control-Max-Age', maxAge)
 
-    if (preflightContinue) {
-      next?.()
+    const method = req.method && req.method.toUpperCase && req.method.toUpperCase()
+    if (method === 'OPTIONS') {
```
talentlessguy, line 64: Use `?.` (optional chaining) operator instead

talentlessguy (changes requested):

packages/cors/src/index.ts
aldy505: done

talentlessguy (approved):

talentlessguy:
@all-contributors add @aldy505 for bug

PR #240, opened 2021-02-26
Fix dependency count

lakgani:
Home page mentioned 4 of 6 dependencies are of tinyhttp's but when I looked at the dependencies on npm I see 5 of them are of tinyhttp's. This tiny PR is to fix that
image


talentlessguy:
@all-contributors add @lakgani for docs

PR #244, opened 2021-04-13
fix: res.send(buffer)

khmm12:
Before it didn't send anything except headers and didn't close a connection.

### DIFF
--- a/packages/send/src/send.ts
+++ b/packages/send/src/send.ts
@@ -22,18 +22,18 @@ export const send = <Request extends Req = Req, Response extends Res = Res>(req:
 ): Response => {
   let bodyToSend = body
 
-  // in case of object - turn it to json
-  if (typeof body === 'object' && body !== null) {
+  if (Buffer.isBuffer(body)) {
+    bodyToSend = body
+  } else if (typeof body === 'object' && body !== null) {
+    // in case of object - turn it to json
     bodyToSend = JSON.stringify(body, null, 2)
-  } else {
-    if (typeof body === 'string') {
-      // reflect this in content-type
-      const type = res.getHeader('Content-Type')
+  } else if (typeof body === 'string') {
+    // reflect this in content-type
+    const type = res.getHeader('Content-Type')
 
-      if (type && typeof type === 'string') {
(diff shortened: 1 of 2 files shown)


packages/send/src/send.ts
```diff
@@ -22,17 +22,19 @@ export const send = <Request extends Req = Req, Response extends Res = Res>(req:
...
-      // reflect this in content-type
-      const type = res.getHeader('Content-Type')
+  } else if (typeof body === 'string') {
+    // reflect this in content-type
+    const type = res.getHeader('Content-Type')
 
-      if (type && typeof type === 'string') {
-        res.setHeader('Content-Type', setCharset(type, 'utf-8'))
-      } else res.setHeader('Content-Type', setCharset('text/html', 'utf-8'))
+    if (type && typeof type === 'string') {
```
talentlessguy, line 34: not sure what's the point in adding line breaks?

packages/send/src/send.ts
```diff
@@ -64,7 +66,10 @@ export const send = <Request extends Req = Req, Response extends Res = Res>(req:
       return
     } else if (Buffer.isBuffer(body)) {
       if (!res.getHeader('Content-Type')) res.setHeader('content-type', 'application/octet-stream')
-    } else encoding ? json(res)(bodyToSend, encoding) : json(res)(bodyToSend)
+      res.end(bodyToSend)
+    } else {
```
talentlessguy, line 70: here as well

packages/send/src/send.ts
khmm12: ok, since it's matter of taste I reverted these lines change.

packages/send/src/send.ts
khmm12: reverted as well

packages/send/src/send.ts
khmm12: to my surprise prettier doesn't add line breaks on multiline blocks yet 🙂

talentlessguy:
LGTM, merging

talentlessguy:
@all-contributors add @khmm12 for bug

PR #268, opened 2021-06-18
example: add socket-io example

wakeupmh:
Just using this superb lib with socket.io, so I made a simple example for it 😊


talentlessguy:
but there's already a ws example

I know those are different but I don't think it's worth adding socketio example since it's quite old (and outdated imo)

PR #271, opened 2021-06-20
Fix statusEmoji import in logger

tbhmens:


talentlessguy:
thanks, i remember it's a cjs module so there should be default import

talentlessguy:
could you please pull from master?

PR #275, opened 2021-06-22
added @tinyhttp/swagger

massivefermion:
Hi
a while back I wanted to use tinyhttp in a project at my work, but because I needed swagger documentation and I'm too lazy to actually write it myself, I used fastify instead which does have automatic swagger doc generation.
But because I liked tinyhttp and I also wanted to start contributing to open source, I decided to do it myself. So here we are.
The way it works, is that you give each route that you want to appear in your documentation, a middleware and pass validation data to it in the style of `fastest-validator`(that's what you decided to use for validation, right?). Currently it ignores anything other than `type`, `optional` and `items` in case `type` is `array`.
The object can have `headers`, `params`, `query` and `body`. Here is an example:

```javascript
import { addToDocs, generateDocs } from '@tinyhttp/swagger'

const schema = {
  id: { type: 'number', positive: true, integer: true },
  name: { type: 'string', min: 3, max: 255 },
  status: 'boolean' 
}

const app = new App()

app
  .post('/', 
    addToDocs({ headers: {authorization: 'string'}, params: { docId: 'number' }, body: schema }),
    (req, res) => {
      res.status(200).send('done')
  })


writeFileSync('docs.json', JSON.stringify(generateDocs(app)), { encoding: 'utf-8' })
app.listen(3000)
```

Currently it only works if you call `generateDocs` after all the routes are registered. I also had problem using `swagger-ui-express` to serve it. But I'm not sure I have more time  to figure it out. So currently you have to serve the doc manually.
I know the api is a bit awkward but I'm just trying to get the ball rolling so maybe others can help me make it better.
Hope this will be helpful.

### DIFF
--- a/packages/swagger/package.json
+++ b/packages/swagger/package.json
@@ -1,0 +1,22 @@
+{
+  "name": "swagger",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs"
+    },
+    "./package.json": "./package.json",
+    "./": "./"
+  },
+  "devDependencies": {
+    "@types/swagger-ui-express": "^4.1.2",
+    "ts-node": "^10.0.0",
+    "typescript": "^4.3.2"
+  },
+  "dependencies": {
+    "@tinyhttp/app": "workspace:*",
+    "@tinyhttp/router": "workspace:*",
+    "json-format": "^1.0.1",
+    "swagger-ui-express": "^4.1.6"
+  }
+}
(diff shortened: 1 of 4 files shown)


talentlessguy:
Hi!

Damn, I wouldn't expect that people would use my project at work, that's amazing to hear! Regarding this PR, it looks really cool and would be useful as an addition to the middleware collection. However, I'm currently in process of moving middleware away from monorepo to separate repos (less maintenance burden, faster CI).

Would you like me to invite you to tinyhttp org and then I create a swagger repo for you?

it'd be smth like https://github.com/tinyhttp/swagger

massivefermion:
that's great. ok.

talentlessguy:
sent you an invite

massivefermion:
Thanks

talentlessguy:
closing because of #283

PR #283, opened 2021-06-29
added swagger example

massivefermion:


talentlessguy:
@all-contributors add @massivefermion for plugin and example

Issue #311, opened 2021-09-03
Not compatible with Node.js v12 because optional chaining in published code


PabloSzx:
**Describe the bug**

Code in published library v2.0.0 is not compatible with Node v12 because it's not converting the optional chaining for it to be compatible with Node v12

![Screenshot from 2021-09-03 17-50-48](https://user-images.githubusercontent.com/8672915/132068485-da90a4c1-1a17-409a-8459-0161754b7f32.png)

**Expected behavior**

tinyhttp to actually support Node.js v12 as the announcements mentioned it

**Versions**

- `node`: 12.22.6
- `@tinyhttp/app`: 2.0.0

talentlessguy:
what's the minor and patch version of node you use? is it higher than 12.4?

PabloSzx:
12.22.6 the latest v12, you can see in the screenshot the optional chaining

talentlessguy:
weird, TS should transpile this

not sure why it's included in the dist

I probably messed up somewhere in configs

will take a look tomorrow

PabloSzx:
@talentlessguy pretty sure this is not normal
![image](https://user-images.githubusercontent.com/8672915/132367045-ec21d097-4848-44b6-b286-f16788edfa12.png)

![image](https://user-images.githubusercontent.com/8672915/132367073-1bedb54a-ad8c-4944-a7f2-dacc0102d24a.png)

PR #313, opened 2021-09-10
chore: fix hanging build

aldy505:
Resolves this problem: https://github.com/tinyhttp/tinyhttp/issues/311#issuecomment-914381040

because of this problem: https://github.com/rollup/plugins/issues/983

and I still don't get why you didn't commit the lock file for the main repo. I think that caused Github Actions to hangs because typescript will search for its' newest version instead of the one that pointed on the package.json. See [this](https://github.com/npm/node-semver#advanced-range-syntax) and [this](https://stackoverflow.com/a/22149167)

let me know if you think of something else.


talentlessguy:
@all-contributors add @aldy505 for fix

PR #337, opened 2021-11-25
fix: omit optional URL params when not supplied in request

### LINKED ISSUE #331: Omitted optional param value is `"undefined"`
**Describe the bug**

When an optional route param (i.e. `/foo/:bar?`) is not supplied for a particular request, the `req.params` object will still contain it, with a string value of `"undefined"`.

This would mean that i.e. `if (req.params.bar)` would unexpectedly always be truthy, even when the param was not supplied.

**To Reproduce**

Simple app with optional route param `bar`:

```js
import { App } from '@tinyhttp/app';
const app = new App();

app.get('/foo/:bar?', (req, res) => {
	console.log(req.params.bar);
	console.log(typeof req.params.bar);
});

app.listen(3000);
```

Visiting `http://localhost:3000/foo/bar` will correctly log:

```
bar
string
```

However, when you omit the param, `http://localhost:3000/foo` it will log:

```
undefined
string
```

**Expected behavior**

The value of the omitted param should not be a string of `"undefined"`. Instead, it should either be a literal `undefined` or `null` - or maybe the `bar` key should not be contained in `request.params` altogether.

**Versions**

- `node`: 14.18.1
- `@tinyhttp/app`: 2.0.11

groenroos:
When optional URL params (i.e. `/foo/:bar?`) are not supplied in the request, the corresponding key should also not be present in the `getURLParams` result.

Closes #331

### DIFF
--- a/packages/url/src/index.ts
+++ b/packages/url/src/index.ts
@@ -12,7 +12,9 @@ export const getURLParams = ({ pattern, keys }: Regex, reqUrl = '/'): URLParams 
 
   if (matches && typeof keys !== 'boolean')
     for (let i = 0; i < keys.length; i++) {
-      params[keys[i]] = decodeURIComponent(matches[i + 1])
+      if (matches[i + 1]) {
+        params[keys[i]] = decodeURIComponent(matches[i + 1])
+      }
     }
 
   return params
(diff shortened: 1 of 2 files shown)


talentlessguy:
@all-contributors add @groenroos for code

PR #352, opened 2022-03-25
Fix: Issue #351 default server host compatibility with IPv6

### LINKED ISSUE #351: Default server host not compatibile with IPv6
**Describe the bug**

If you create a basic server and don't specify an address to bind to, the server will bind to 0.0.0.0. On an OS with both IPv4 and IPv6, this may mean it will reject requests to `localhost`.

**To Reproduce**

1. install and run the [basic example](https://github.com/tinyhttp/tinyhttp/tree/master/examples/basic)
2. open http://localhost:3000 in a browser, see "Hello World" message.
3. http://127.0.0.1:3000/ works too.
4. open `http://[::1]:3000/` (IPv6 loopback address) - fails

**Expected behavior**

You should see the same "Hello World" in all cases. A server without a specified host should listen on all available interfaces.

**Workaround:** change [line 25 of index.js](https://github.com/tinyhttp/tinyhttp/blob/master/examples/basic/index.js#L25) to this:

```js
.listen(3000, () => console.log(`Listening on http://localhost:3000`), "::")
``` 

**Versions**

- `node`: 17.6.0
- `@tinyhttp/app`: 2.0.19

**Additional context**

I personally found this while trying to use tinyhttp to power the API backend for a Vite frontend, and doing local testing using Vite's [built-in proxy](https://vitejs.dev/config/#server-proxy). The frontend was running on http://localhost:3000 and the backend was using http://localhost:3001. Accessing both services through a browser was fine, and also running the whole stack inside of docker compose was OK too. But on my MacBook (macOS 12.2.1) using Node v17.6.0, attempts to proxy to the backend resulted in this error:

```text
Error: connect ECONNREFUSED ::1:3001
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1158:16)
```

...until I changed the `app.listen` line in my server to add the last argument with the value `"::"`.

```js
app.listen(PORT, () => console.log("Started on http://localhost:" + PORT), "::");
```

It turns out that a blank string `""` also works. On my machine at least, IPv6 seems to have higher priority than IPv4. AFAICT, this behavior appears to have been changed in GH-86 to fix #85. According to [the relevant Node documentation for server.listen](https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback):

> If `host` is omitted, the server will accept connections on the [unspecified IPv6 address](https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address) (::) when IPv6 is available, or the [unspecified IPv4 address](https://en.wikipedia.org/wiki/0.0.0.0) (0.0.0.0) otherwise.
> In most operating systems, listening to the [unspecified IPv6 address](https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address) (::) may cause the `net.Server` to also listen on the [unspecified IPv4 address](https://en.wikipedia.org/wiki/0.0.0.0) (0.0.0.0).

This suggests that in addition to the workaround above (that is, using host value `"::"` instead of `"0.0.0.0"`), perhaps omitting the host by making the default host an empty string is the right call, and let Node sort it out. 

The reasoning behind dual-stack interface binding is discussed further in nodejs/node#9390.

By the way, great library and documentation. Thanks for your work on this over the years.

Barnabas:
Closes #351 by un-defining the default server host name and adds a new test which passes on dual-stack machines.

Sadly, this may not be testable on GitHub due to actions/virtual-environments#668.

### DIFF
--- a/packages/app/src/app.ts
+++ b/packages/app/src/app.ts
@@ -340,7 +340,7 @@ export class App<
    * @param Server callback after server starts listening
    * @param host server listening host
    */
-  listen(port?: number, cb?: () => void, host = '0.0.0.0'): Server {
+  listen(port?: number, cb?: () => void, host?: string): Server {
     return createServer().on('request', this.attach).listen(port, host, cb)
   }
 }
--- a/tests/core/request.test.ts
+++ b/tests/core/request.test.ts
@@ -47,27 +47,33 @@ describe('Request properties', () => {
   })
 
   describe('Network extensions', () => {
+    const ipHandler = (req, res) => {
+      res.json({
+        ip: req.ip,
+        ips: req.ips
+      })
+    }
+    const options = {
+      settings: {
+        networkExtensions: true
+      }
+    }
     it('IPv4 req.ip & req.ips is being parsed properly', async () => {
-      const { fetch } = InitAppAndTest(
-        (req, res) => {
-          res.json({
-            ip: req.ip,
-            ips: req.ips
-          })
-        },
-        '/',
-        'GET',
-        {
-          settings: {
-            networkExtensions: true
-          }
-        }
-      )
+      const { fetch } = InitAppAndTest(ipHandler, '/', 'GET', options)
 
       const agent = new Agent({ family: 4 }) // ensure IPv4 only
       await fetch('/', { agent }).expect(200, {
         ip: '127.0.0.1',
         ips: ['::ffff:127.0.0.1']
+      })
+    })
+    it('IPv6 req.ip & req.ips is being parsed properly', async () => {
+      const { fetch } = InitAppAndTest(ipHandler, '/', 'GET', options)
+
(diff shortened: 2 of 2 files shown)


talentlessguy:
CI seems to fail

probably worth skipping IPv6 tests?

talentlessguy:
@aarontravass I think GitHub actions don't support IPv6 so I guess I'll just manually remove the failing test and will merge this

aarontravass:
> @aarontravass I think GitHub actions don't support IPv6 so I guess I'll just manually remove the failing test and will merge this

How about testing ipv6 only if run on non github ci/cd runner? Actions has some default variables that you can utilize - https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

talentlessguy:
@aarontravass good idea, feel free to extend this PR and update the tests

PR #356, opened 2022-04-24
feat: Implement res.locals behavior (#355)

### LINKED ISSUE #355: Utilize res.locals as expected
**Is your feature request related to a problem? Please describe.**
Unsure if this is a bug report or a feature request, I've gone with the latter as this seems to be unimplemented behavior.

In express, res.locals are a per-request equivalent to app.locals, and the property is mentioned here in tinyhttp:

https://github.com/tinyhttp/tinyhttp/blob/072e75fea7eed2ff341c2b5ca25d8410c8762e7f/packages/app/src/response.ts#L47

However, res.locals doesn't seem to be implemented, as it doesn't seem to be used in [renderTemplate](https://github.com/tinyhttp/tinyhttp/blob/072e75fea7eed2ff341c2b5ca25d8410c8762e7f/packages/app/src/response.ts#L7-L20), and they won't be available within templates:

```js
import { App } from "@tinyhttp/app";
import { renderFile as eta } from "eta";
const app = new App();
app.engine("eta", eta);
app.use(function(req, res, next) {
    res.locals.hello = "hello world!";
    next();
});
app.get((req, res) => res.render("index.eta"));
app.listen(3000, () => console.log("Listening on port 3000"));
```
views/index.eta
```html
<body>
  <p><%= it.hello %></p>
</body>
```

**Describe the solution you'd like**
Matching express's res.locals behavior, described here: https://expressjs.com/en/api.html#res.locals

**Describe alternatives you've considered**
I could manually use res.locals with res.render or use a function to automate this, but it would be more convienent to have res.locals's expected behavior, as doing so somewhat defeats the point of res.locals
```js
app.get((req, res) => res.render("index.eta", { hello: res.locals.hello }));
```

06000208:
Closes #355

If there's a better way to do this or if this should be done somewhere else in the code lmk

### DIFF
--- a/packages/app/src/response.ts
+++ b/packages/app/src/response.ts
@@ -9,7 +9,7 @@ export const renderTemplate =
   (file: string, data?: Record<string, any>, options?: TemplateEngineOptions<O>): Response => {
     app.render(
       file,
-      data,
+      data ? { ...data, ...res.locals } : data,
       (err: unknown, html: unknown) => {
         if (err) throw err
         res.send(html)


talentlessguy:
could you add a test please?

Issue #358, opened 2022-05-28
App is seen with type "any" when imported


jgoux:
Hello 👋, I'm building a small demo repo including full ESM support + Typescript and I have an issue with tinyhttp.

**Describe the bug**

When importing `App` from `@tinyhttp/app`, the type is `any`.

**To Reproduce**

1. Clone https://github.com/jgoux/acme
2. Run `yarn`
3. Go to https://github.com/jgoux/acme/blob/main/apps/api/src/index.ts#L7

**Expected behavior**

The `App` should be strongly typed, not as `any`

**Versions**

- `node`: 16.15.0
- `@tinyhttp/app`: 2.0.20

**Additional context**

<img width="800" alt="image" src="https://user-images.githubusercontent.com/1443499/170835650-9196b56e-1ff4-480d-95ea-fcaee131da90.png">

jgoux:
@talentlessguy Sorry maybe my monorepo starter isn't a great repro.

Here is one more scoped with the bare minimal files showing the issue: https://github.com/jgoux/tinyhttp-repo

Maybe it's something due to TypeScript 4.7 `"module": "NodeNext"` which introduced changes in how ESM modules are seen?

`@tinyhttp/app` is the only ESM module that causes this issue in my monorepo example.

talentlessguy:
aight, could you submit a PR then to fix this?

jgoux:
Ok I have some news, I got help on TypeScript's Discord.

The main issue is that in ESM, local files should be imported with their extension explicitly (even in TypeScript):
- https://www.typescriptlang.org/docs/handbook/2/modules.html#modules-in-typescript
- https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-make-my-typescript-project-output-esm

It's not enforced right now because of `"moduleResolution": "Node"` in https://github.com/tinyhttp/tinyhttp/blob/master/tsconfig.json#L15

I think the best course of action would be to bump TS to 4.7 and use `"moduleResolution": "NodeNext"`.
Also, all the imports targeting local files will need to include their full path, including the `.js` extension.

PR #359, opened 2022-05-30
fix(esm): add file extension to imports and exports

jgoux:
Resolves https://github.com/tinyhttp/tinyhttp/issues/358

I didn't bump TypeScript because adding the extensions was enough. 👍 (I tried along with targeting `"module": "Node16"` but there were failures with jest, and I think it would need another PR) 

I just run those two regexes in VSCode:
image
image

As the tests are green, I assume I didn't forget anything! ✅


jest.config.mjs
```diff
@@ -13,6 +13,9 @@ export default {
     '^.+\\.(t|j)sx?$': 'ts-jest'
   },
   testRegex: '(/tests/.*|\\.(test))\\.(ts|tsx|js)$',
+  moduleNameMapper: {
+    '^(\\.{1,2}/.*)\\.js$': '$1'
+  },
```
jgoux, line 18: I had to add this because files imported with the `.js` extension weren't seen by jest, as jest still have a limited support for ESM. I followed `ts-jest`: https://kulshekhar.github.io/ts-jest/docs/guides/esm-support/#examples

crutchcorn: Just a small 2 cents on the Jest ESM interop for the maintainers of TinyHTTP:

Jest caused me all kinds of headaches with ESM. With [PlopJS](https://github.com/plopjs/plop) we migrated everything to [Vitest](https://vitest.dev/) in ~5 minutes (it has a Jest compat mode, so no need to rewrite tests) and have seemingly had no related issues that we did with Jest.

Happy to send a follow-up PR if this sounds interesting to the TinyHTTP maintainers.

talentlessguy: I'd be down

Issue #360, opened 2022-08-29
res.sendFile should respect content type when set


raymcdermott:
**Is your feature request related to a problem? Please describe.**
res.sendFile does not get the content type correct 100% of the time and yet it does not respect content type when it is set on the response.

**Describe the solution you'd like**
res.sendFile should respect content type when it is set on the response.

**Describe alternatives you've considered**
I currently read the files locally and set the content. This is more code and I would rather not maintain it.

**Additional context**
Firefox gives XML errors when content types don't match its expectations - even if those expectations are wrong. It would be nice to be able to set the content-type as described here to avoid such noise.

https://stackoverflow.com/questions/51000009/i-keep-getting-this-error-xml-parsing-error-syntax-error-but-still-the-website

talentlessguy:
Could you please submit a PR with the bug fix?

rokaicker:
Hey @talentlessguy, if nobody is looking into this, is it okay if I give it a shot? 

But I am wondering, is it known under which circumstances it doesn't get the correct content-type? Ex. specific file types, etc.

talentlessguy:
@rokaicker sure

I don't know myself but it needs checking

PR #363, opened 2022-10-12
fix: prevent res.sendFile() from overwriting 'Content-Type' header if previously set

rokaicker:
This PR aims to resolve [Issue#360](https://github.com/tinyhttp/tinyhttp/issues/360).

In sendFile.ts, I added a check to see if the "Content-Type" header had been previously set, and it will only pull the MIME type from the file if the header has not been set.

I then added an extra test inside send.test.ts, where it sets the "Content-Type" header to text/markdown for a text/plain file, to see if it persists. 

Please let me know if any advice/ changes required. 

Thanks!

### DIFF
--- a/packages/send/src/sendFile.ts
+++ b/packages/send/src/sendFile.ts
@@ -34,7 +34,7 @@ export type Caching = Partial<{
 
 type Req = Pick<I, 'headers'>
 
-type Res = Pick<S, 'setHeader' | 'statusCode' | 'writeHead'> & NodeJS.WritableStream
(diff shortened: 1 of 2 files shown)


rokaicker:
@talentlessguy Hey, I just noticed the coverage check wasn't successful, because coverage decreased by 0.01%, I was wondering is this acceptable? Thanks!

talentlessguy:
it seems to upload coverage of `dist` files which shouldn't happen

PR #369, opened 2022-12-10
Update onError.ts

sachaw:


talentlessguy:
do type imports need an extension?

sachaw:
Yes they do, typescript will throw an error as it doesn't differentiate about imports vs type imports in this case

talentlessguy:
aight then merging

PR #378, opened 2023-04-21
Correct type declaration

bullno1:
* App.locals should be `Record<string any>` because the app doesn't really care about the values, it just pass them to functions. They can even be `<string, unknown>`.
* Callback for template rendering function should be `(error: Error | null, html: unknown) => void` because error can be null.


talentlessguy:
thanks for the PR, gonna run CI and make suer it's cool and then i'll merge

Issue #377, opened 2023-02-24
Multiple path arguments (as array) to route


kevteljeur:
**Is your feature request related to a problem? Please describe.**
Support for arrays of paths as the path argument to a route

**Describe the solution you'd like**
Currently only single paths are supported. ExpressJS supports arrays of paths to a route, so that multiple paths can resolve to the same route without needing each route to be specified separately.

**Describe alternatives you've considered**
The alternative is as now to create separate routes for each alternative path.

**Additional context**
```
app.get( [ '/child', '/other-child' ], ( req, res ) => {
	res.send( `${req.method || 'GET' } request to homepage` )
} )
```
vs 

```
app.get( '/child', ( req, res ) => {
	res.send( `${req.method || 'GET' } request to homepage` )
} )

app.get( '/other-child', ( req, res ) => {
	res.send( `${req.method || 'GET'} request to homepage` )
} )
```

talentlessguy:
when I tried to implement it I had TS type conflicts, so PR with this feature is welcome

jasonghent98:
Hi, is this issue still open and relevant? Would like to help resolve if so

talentlessguy:
@jasonghent98 yes it's still relevant, a PR with this enhancement is very welcome

kevteljeur:
Thank you!

PR #379, opened 2023-05-06
feat: Issue #377 added the ability to accept multiple paths 

aarontravass:
As mentioned in Issue #377, methods can now take an array of paths.


aarontravass:
Can we have `pnpm format:fix` as part of the pre commit hook?

talentlessguy:
yeah but you should add it in a separate PR

.gitpod.yml
talentlessguy, line 1: please delete this file

aarontravass:
> yeah but you should add it in a separate PR

Yes, will do that.

packages/router/src/index.ts
```diff
@@ -206,13 +206,27 @@ export class Router<App extends Router = any, Req = any, Res = any> {
...
     return (...args: RouterMethodParams<Req, Res>): this => {
       const handlers = args.slice(1).flat() as Handler<Req, Res>[]
-      pushMiddleware<Req, Res>(this.middleware)({
-        path: args[0],
-        handler: handlers[0],
-        handlers: handlers.slice(1),
-        method,
-        type: 'route'
-      })
+      if (typeof args[0] == 'object') {
```
talentlessguy, line 209: I would replace this check with `Array.isArray`

tests/modules/router.test.ts
```diff
@@ -373,6 +373,13 @@ describe('Testing HTTP methods', () => {
...
     router.all('/', () => void 0)
 
+    expect(router.middleware[0].type).toBe('route')
+  })
+  it('app.get with array of paths should set GET as HTTP method', () => {
+    const router = new Router()
+
+    router.get(['/', '/test'], () => void 0)
+
     expect(router.middleware[0].type).toBe('route')
```
talentlessguy, line 383: I think it'd make sense to assert `middleware.length` instead

talentlessguy (changes requested):

tests/modules/router.test.ts
aarontravass: Is this how the length is checked? 
`expect(router.middleware.length).toHaveLength(2)`

aarontravass: Fixed in commit ad67943ae61c9e428f7bf41ffd424ca2b2d94ddd

aarontravass:
Anything more to be addressed?

talentlessguy:
nope, I'll merge it and release it within next two days

PR #383, opened 2023-05-07
chore: added caching example

aarontravass:
Added a caching example which uses cache-manager for the backend.


examples/caching/weather.js
```diff
@@ -0,0 +1,8 @@
+import axios from 'axios'
```
talentlessguy, line 1: there is no need to introduce an extra dependency. you can use `https` module:

```js
import https from 'node:https'

https
  .get(`https://reqres.in/api/users`, resp => {
    let data = "";

    // A chunk of data has been recieved.
    resp.on("data", chunk => {
      data += chunk;
    });

    // The whole response has been received. Print out the result.
    resp.on("end", () => {
      let url = JSON.parse(data).message;
      console.log(url);      
    });
  })
  .on("error", err => {
    console.log("Error: " + err.message);
  })
```

talentlessguy (changes requested):

examples/caching/weather.js
aldy505: Or just use global `fetch` https://nodejs.org/api/globals.html#fetch

aarontravass: For some reason fetch was not available. Probably because my node version was old.

aarontravass:
> is it possible to replace it with lru-send? cuz it's also a caching solution

let me give it a try

aarontravass:
> is it possible to replace it with lru-send? cuz it's also a caching solution

done

examples/caching/weather.js
```diff
@@ -0,0 +1,6 @@
+const URL = 'https://goweather.herokuapp.com/weather/'
```
talentlessguy, line 1: Can you please move this to the file with the rest of the code plz? It's ~25 LOC anyway

talentlessguy (changes requested):

aldy505 (approved):

PR #388, opened 2023-05-19
Router and dotenv coverage is 100%

aarontravass:
Completes #92 router and dotenv coverage

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,5 +5,6 @@
   "prettier.enable": true,
   "[typescript]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
-  }
+  },
+  "editor.tabSize": 2
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,6 +5,5 @@
   "prettier.enable": true,
   "[typescript]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
-  },
-  "editor.tabSize": 2
+  }
 }
--- a/packages/router/src/index.ts
+++ b/packages/router/src/index.ts
@@ -260,21 +260,6 @@ export class Router<App extends Router = any, Req = any, Res = any> {
   }
 
   /**
-   * Return the app's absolute pathname
-   * based on the parent(s) that have
-   * mounted it.
-   *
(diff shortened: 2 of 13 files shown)


.vscode/settings.json
talentlessguy, line 1: tabSize is handled by prettier

packages/router/src/index.ts
```diff
@@ -259,21 +259,6 @@ export class Router<App extends Router = any, Req = any, Res = any> {
     return this
   }
 
-  /**
```
talentlessguy, line 262: why move it away from router?

talentlessguy (changes requested):

packages/router/src/index.ts
aarontravass: coverage showed the left branch was never reached
```typescript
this.parent ? this.parent.path() + this.mountpath : ''
```

Which is true because the Router by itself has no parent. Instead, the app has a parent and thus makes sense if moved to App.

talentlessguy: hm maybe we should assign the parent in the router in this case

aarontravass: it won't if the Router by itself is initialized. For example
```typescript
const router = new Router()
router.parent // will always be null
router.path() // will be /
```

aarontravass: what I mean is we can't mount the Router. So the parent of the Router can never have a non null value.

talentlessguy: alright got it, then moving it to app makes sense

tests/core/app.test.ts
```diff
@@ -304,6 +304,24 @@ describe('App methods', () => {
 
     await makeFetch(server)('/', { method: 'POST' }).expect(200, 'POST request')
   })
+  it('app.find should return all metadata information for a given route', async () => {
```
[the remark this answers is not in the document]
aarontravass: hmmm you can probably make it a private method to avoid being accessed outside of the `App` class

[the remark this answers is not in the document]
aarontravass: nevermind, I just saw your commit

talentlessguy (approved):

PR #394, opened 2023-06-24
coverage is 100%

aarontravass:


packages/app/src/app.ts
```diff
@@ -181,8 +180,7 @@ export class App<
     if (options._locals) locals = { ...locals, ...options._locals }
 
     if (!file.endsWith(`.${options.ext}`)) file = `${file}.${options.ext}`
-
-    const dest = options.viewsFolder ? path.join(options.viewsFolder, file) : file
+    const dest = path.join(options.viewsFolder, file)
```
talentlessguy, line 183: why remove a ternary? it might be that views folder is set to `''`

packages/app/src/request.ts
```diff
@@ -18,8 +18,10 @@ export { getURLParams } from '@tinyhttp/req'
 const trustRemoteAddress = ({ socket }: Pick<Request, 'headers' | 'socket'>) => {
   const val = socket.remoteAddress
 
-  if (typeof val === 'string') return compile(val.split(',').map((x) => x.trim()))
-
+  if (typeof val === 'string') {
```
talentlessguy, line 21: actually here `val` is always string so we can remove the unreachable code

packages/app/src/request.ts
```diff
@@ -29,9 +31,10 @@ export const getRouteFromApp = ({ middleware }: App, h: Handler<Request, Respons
 export const getProtocol = (req: Request): Protocol => {
   const proto = `http${req.secure ? 's' : ''}`
 
+  /* c8 ignore next */
   if (!trustRemoteAddress(req)) return proto
 
-  const header = (req.headers['X-Forwarded-Proto'] as string) || proto
+  const header = (req.get('X-Forwarded-Proto') as string) || proto
```
talentlessguy, line 37: let's keep `req.headers`

packages/app/src/request.ts
```diff
@@ -40,9 +43,8 @@ export const getProtocol = (req: Request): Protocol => {
 
 export const getHostname = (req: Request): string | undefined => {
   let host: string = req.get('X-Forwarded-Host') as string
-
+  /* c8 ignore next */
```
talentlessguy, line 46: I think we should test in environment when there's no such header defined and address can't be trusted

packages/content-disposition/src/index.ts
```diff
@@ -111,7 +110,7 @@ function createParams(filename?: string, fallback?: string | boolean) {
   }
 
   // set filename parameter
-  if (isQuotedString || hasFallback) {
+  if (hasFallback || isQuotedString) {
```
talentlessguy, line 113: why change order here?

packages/jsonp/src/index.ts
```diff
@@ -25,8 +25,6 @@ function stringify(
           return '\\u003e'
         case 0x26:
           return '\\u0026'
-        default:
```
talentlessguy, line 28: why remove default case? it should fallback if switch falls through

tests/core/app.test.ts
```diff
@@ -53,6 +53,22 @@ describe('Testing App', () => {
 
     await fetch('/').expect(500, 'Ouch, you hurt me on / page.')
   })
+  it('Default onError with testing', async () => {
+    process.env = {
+      ...process.env,
```
talentlessguy, line 58: use `vi.stubEnv` to mock env variables

packages/app/src/app.ts
aarontravass: that is not possible due to 
```ts
options.viewsFolder = options.viewsFolder || this.settings.views || `${process.cwd()}/views`
```
always having a fall back to at least `/views` (assuming `process.cwd()` is '')

packages/app/src/request.ts
aarontravass: can `val` be null?

packages/app/src/request.ts
aarontravass: `req.headers` was null when I supplied it in this test
https://github.com/aarontravass/tinyhttp/blob/670985580bf4fd0db48d4bfa3c9801813e5be23c/tests/core/request.test.ts#L150

packages/jsonp/src/index.ts
aarontravass: The regex `/[<>&]/g` won't allow any other char. Hence, it won't reach the default.

packages/app/src/request.ts
aarontravass: [`trustRemoteAddress` always returns a fn](https://github.com/tinyhttp/tinyhttp/blob/c5d4a90b3c2cf8ded8c5c1f281dd719ebf8cfcc5/packages/proxy-addr/src/index.ts#L69) and hence it is an unreachable branch

aarontravass:
@talentlessguy should we set min coverage for lines, fns, statements and branches to 100%? https://vitest.dev/config/#coverage-lines

packages/app/src/request.ts
talentlessguy: let's eliminate the unreachable code then

packages/app/src/request.ts
talentlessguy: hm this is weird, `req.get` does the same thing as `req.headers[...]`

packages/app/src/request.ts
talentlessguy: I'm not sure, needs digging in Node.js docs

aarontravass: it can be undefined

> The string representation of the remote IP address. For example, '74.125.127.100' or '2001:4860:a005::68'. Value may be undefined if the socket is destroyed (for example, if the client disconnected).

https://nodejs.org/api/net.html#socketremoteaddress

packages/app/src/request.ts
aarontravass: image
yeah. dont know why. here's the screenshot

packages/forwarded/src/index.ts
```diff
@@ -41,7 +28,20 @@ export function parse(header: string): string[] {
   }
 
   // final address
-  if (start !== end) list.push(header.substring(start, end))
 
+  if (start !== end) list.push(header.substring(start, end))
```
talentlessguy, line 32: why c8 ignore here?

btw let's not add ignored coverage, so we don't miss anything

talentlessguy (changes requested):

packages/forwarded/src/index.ts
aarontravass: it's kinda buggy so it thinks the bracket is uncovered. 
Screenshot 2023-08-03 at 14-06-22 Code coverage report for packages_forwarded_src_index ts

packages/send/src/sendFile.ts
```diff
@@ -42,6 +42,7 @@ export const enableCaching = (res: Res, caching: Caching): void => {
   else if (cc && caching.maxAge === 0) cc += ',must-revalidate'
 
   if (cc) res.setHeader('Cache-Control', cc)
+  /* c8 ignore next */
```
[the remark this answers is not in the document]
aarontravass: same issue 
Screenshot 2023-08-03 at 14-10-01 Code coverage report for packages_send_src_sendFile ts

talentlessguy:
@aarontravass we could set min coverage for lines as 80% so it's not as strict

aarontravass:
> @aarontravass we could set min coverage for lines as 80% so it's not as strict

80 is too low tho

aarontravass:
> > > @aarontravass we could set min coverage for lines as 80% so it's not as strict
> > 
> > 
> > 80 is too low tho
> 
> 95 then ig

yeah should be fine

packages/content-disposition/src/index.ts
```diff
@@ -89,9 +88,9 @@ function createParams(filename?: string, fallback?: string | boolean) {
   > = {}
 
   // fallback defaults to true
-  if (fallback === undefined) fallback = true
-
-  if (typeof fallback === 'string' && NON_LATIN1_REGEXP.test(fallback)) {
+  if (!fallback) fallback = true
+  const res = typeof fallback === 'string' && NON_LATIN1_REGEXP.test(fallback)
```
talentlessguy, line 92: why change here?

packages/forwarded/src/index.ts
```diff
@@ -1,25 +1,12 @@
...
 /**
  * Parse the X-Forwarded-For header.
  */
-export function parse(header: string): string[] {
+export const parse = (header: string) => {
   let end = header.length
-  const list: string[] = []
   let start = header.length
-
+  const list: string[] = []
```
talentlessguy, line 9: why moving it?

packages/forwarded/src/index.ts
talentlessguy: it's fine, it's probably something else causing the cov miss

tests/modules/res.test.ts
```diff
@@ -496,3 +553,8 @@ describe('Response extensions', () => {
     })
   })
 })
+describe('util tests', () => {
+  it('should call accept', () => {
```
talentlessguy, line 557: this test description doesn't quite reflect the test itself

talentlessguy (changes requested):

packages/forwarded/src/index.ts
aarontravass: dunno
https://github.com/bcoe/c8/issues/81

talentlessguy:
I've just merged a new view engine - let's close this PR and open smaller PR for increasing coverage - without affecting the source. If source has to be changed - it should be a separate PR

PR #408, opened 2023-09-26
fix: closes #406

### LINKED ISSUE #406: Request `.xhr` misses fetch requests (headers are lowercase)
**Describe the bug**

Currently [`req.xhr`](https://tinyhttp.v1rtl.site/docs#reqxhr) only checks for the existence of an exact match of 'X-Requested-With' in the HTTP Headers. However, many fetch implementations (including native Chrome, node-fetch, etc.) automatically lowercase headers before being sent, so a client can't indicate that it's a xhr request.

I'd expect that the `.xhr` check would check for the `X-Requested-With` AND `x-requested-with` headers.

Relevant code: https://github.com/tinyhttp/tinyhttp/blob/master/packages/req/src/index.ts#L56

**To Reproduce**

You can quickly confirm that the [existing code](https://github.com/tinyhttp/tinyhttp/blob/master/packages/req/src/index.ts#L56) won't work by checking how the `Headers` object automatically transforms header names: `Array.from(new Headers({'X-Requested-With': 'XMLHttpRequest'}))` - it gets converted to lowercase.

Full test example:

```js

// server fixture
const app = new App();
app.get('/isxhr/:agent', function (req, res) {
    const data = {
        agent: req.params.agent,
        xhr: req.xhr,
        titleCase: req.headers['X-Requested-With'],
        lowercase: req.headers['x-requested-with']
    };
    console.debug(data);
    res.json(data);
}).listen(3000, async () => {
    // run client test using native node fetch
    clientTest('node-native').catch(err => console.error(err));
});

// run in browser/node/wherever - note using capitals in header name
async function clientTest(agent = 'default') {
    const res = await fetch(`http://localhost:3000/isxhr/${agent}`, {
        headers: { 'X-Requested-With': 'XMLHttpRequest' }
    });
    const result = await res.json();
    console.assert(result.xhr === true, 'XHR should be true!', result);
    return result;
}

**Expected behavior**

`req.xhr` should be `true` if *either* `X-Requested-With` or `x-requested-with` headers = `XMLHttpRequest`

**Versions**

- `node`: 18.12
- `@tinyhttp/app`: 2.1.X

### PR DESCRIPTION
**Pull request by aarontravass**
closes #406

### DIFF
--- a/packages/req/src/index.ts
+++ b/packages/req/src/index.ts
@@ -53,7 +53,9 @@ export const getFreshOrStale = (
 }
 
 export const checkIfXMLHttpRequest = (req: Pick<Request, 'headers'>): boolean =>
-  req.headers['X-Requested-With'] === 'XMLHttpRequest'
+  !!Object.keys(req.headers).find(
+    (header) => header.toLowerCase() === 'x-requested-with' && req.headers[header] === 'XMLHttpRequest'
+  )
 
 export const reqIs =
   (req: Pick<Request, 'headers'>) =>
--- a/packages/req/src/index.ts
+++ b/packages/req/src/index.ts
@@ -53,9 +53,7 @@ export const getFreshOrStale = (
 }
 
 export const checkIfXMLHttpRequest = (req: Pick<Request, 'headers'>): boolean =>
-  !!Object.keys(req.headers).find(
-    (header) => header.toLowerCase() === 'x-requested-with' && req.headers[header] === 'XMLHttpRequest'
-  )
+  req.headers['x-requested-with'] === 'XMLHttpRequest'
 
 export const reqIs =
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/req/src/index.ts [hunk 1]
```diff
@@ -53,7 +53,9 @@ export const getFreshOrStale = (
 }
 
 export const checkIfXMLHttpRequest = (req: Pick<Request, 'headers'>): boolean =>
-  req.headers['X-Requested-With'] === 'XMLHttpRequest'
```
talentlessguy, line 56: header names are lowercase by default

https://nodejs.org/api/http.html#requestgetheaders

you can index by lowercase and it will work already, no need for a custom lookup

talentlessguy (changes requested):

PR #413, opened 2023-10-08
test: added tests for app, ip-filter, proxy-addr, rate-limit and req

aarontravass:


aarontravass:
@talentlessguy

packages/app/src/app.ts
```diff
@@ -353,9 +351,8 @@ export class App<Req extends Request = Request, Res extends Response = Response>
       try {
         params = regex ? getURLParams(regex, pathname) : {}
       } catch (e) {
-        console.error(e)
         if (e instanceof URIError) return res.sendStatus(400) // Handle malformed URI
-        else throw e
```
talentlessguy, line 358: the error must be logged somewhere, like with `onError`

packages/app/src/request.ts
```diff
@@ -16,16 +16,20 @@ export { getURLParams } from '@tinyhttp/req'
...
+    return compile(val.split(',').map((x) => x.trim()))
+    /* c8 ignore next */
+  }
   return compile(val || [])
 }
 
 export const getProtocol = (req: Request): Protocol => {
   const proto = `http${req.secure ? 's' : ''}`
 
+  /* c8 ignore next */
```
talentlessguy, line 29: don't use c8 ignores

package.json
```diff
@@ -21,6 +21,7 @@
     "c8": "^8.0.1",
     "colorette": "2.0.20",
     "dirname-filename-esm": "1.1.1",
+    "ejs": "^3.1.9",
```
talentlessguy, line 24: why add ejs?

tests/core/app.test.ts
```diff
@@ -909,6 +926,14 @@ describe('Subapps', () => {
     const fetch = makeFetch(server)
     await fetch('/%').expect(400, 'Bad Request')
   })
+  it('Should throw an error when url regex throws an error', async () => {
```
talentlessguy, line 929: sending status 500 is not throwing an error

tests/core/app.test.ts
```diff
@@ -1007,19 +1045,22 @@ describe('Template engines', () => {
...
     describe('errors', () => {
       it('should catch errors', () => {
+        expect.assertions(1)
         const app = new App({
           settings: {
             views: `${process.cwd()}/tests/fixtures`
           }
         })
 
-        class TestView {
```
talentlessguy, line 1016: why change?

tests/core/request.test.ts
```diff
@@ -115,6 +116,40 @@ describe('Request properties', () => {
 
       await fetch('/').expect(200, `subdomains: `)
     })
+    describe('custom imports', () => {
```
talentlessguy, line 119: what does "custom imports" mean? internal functions?

talentlessguy (changes requested):

packages/app/src/request.ts
aarontravass: `trustRemoteAddress` always returns a truthy value. Not sure how to test it

package.json
aarontravass: You were defaulting to `.ejs` before. 

https://github.com/tinyhttp/tinyhttp/blob/bec8ee50024a645ad26663752b5761dd03508576/packages/app/src/app.ts#L173

I'll remove it

tests/core/app.test.ts
aarontravass: same
image

tests/fixtures/views/error.ejs
talentlessguy, line 1: Please remove those files - we test against eta, not ejs

talentlessguy (changes requested):

tests/core/request.test.ts
```diff
@@ -115,6 +116,52 @@ describe('Request properties', () => {
 
       await fetch('/').expect(200, `subdomains: `)
     })
+    describe('internal function tests', () => {
```
talentlessguy, line 119: wait, how are they internal if if `getSubdomiains` is used by `req.subdomains`?

tests/core/request.test.ts
```diff
@@ -115,6 +116,52 @@ describe('Request properties', () => {
...
+      it('should test getSubdomains when host is null', async () => {
```
talentlessguy, line 120: this is a vague description, "should test". what does it mean?

talentlessguy (changes requested):

tests/core/request.test.ts
aarontravass: Those functions are not directly expected to be used by the end user.

packages/app/src/request.ts
```diff
@@ -16,16 +16,16 @@ export { getURLParams } from '@tinyhttp/req'
 
 const trustRemoteAddress = ({ socket }: Pick<Request, 'headers' | 'socket'>) => {
   const val = socket.remoteAddress
-  if (typeof val === 'string') return compile(val.split(',').map((x) => x.trim()))
-  return compile(val || [])
+  if (typeof val !== 'string') return compile(val || [])
```
talentlessguy, line 19: why swap it? it's the same logic

tests/core/request.test.ts
```diff
@@ -63,12 +64,12 @@ describe('Request properties', () => {
       })
 
       await fetch('/s/t/u/a1/b/c').expect(200, {
-        url: '/a1/b/c',
+        url: '/s/t/u/a1/b/c',
```
talentlessguy, line 67: why is this test changed?

talentlessguy (changes requested):

packages/app/src/request.ts
aarontravass: `socket.remoteAddress` is always a string so it would never reach line 20

packages/app/src/app.ts
```diff
@@ -338,29 +340,28 @@ export class App<Req extends Request = Request, Res extends Response = Response>
...
     mw.push({
       handler: this.noMatchHandler,
       type: 'mw',
       path: '/'
     })
-
     const handle = (mw: Middleware) => async (req: Req, res: Res, next?: NextFunction) => {
       const { path, handler, regex } = mw
 
+      req.url = reqUrlCopy // reset req.url since it is changed in the handle fn
```
talentlessguy, line 351: why do we reset req.url here?

packages/app/src/request.ts
talentlessguy: then return it directly

packages/app/src/request.ts
```diff
@@ -34,9 +34,7 @@ export const getProtocol = (req: Request): Protocol => {
 
 export const getHostname = (req: Request): string | undefined => {
   let host: string = req.get('X-Forwarded-Host') as string
-
-  if (!host || !trustRemoteAddress(req)) host = req.get('Host') as string
```
talentlessguy, line 38: why did you remove it? trustRemoteAddress may return false, if remote address is different from Host header

tests/core/request.test.ts
talentlessguy: they're not exposed in the app package but you still can use them from `@tinyhttp/req`

packages/app/src/request.ts
aarontravass: looks like it's a string https://nodejs.org/docs/latest/api/net.html#socketremoteaddress

packages/app/src/app.ts
aarontravass: If you have multiple middlewares for 1 path, the `req.url` is modified as per

https://github.com/tinyhttp/tinyhttp/blob/7bf29536a0e5574a82768d87a0b57a34bf217fa5/packages/app/src/app.ts#L376

and hence, every single middleware after the first receives the altered `req.url`.

packages/app/src/request.ts
aarontravass: `trustRemoteAddress` will always return a not null value

https://github.com/tinyhttp/tinyhttp/blob/7bf29536a0e5574a82768d87a0b57a34bf217fa5/packages/proxy-addr/src/index.ts#L88

packages/app/src/app.ts
talentlessguy: req.url is supposed to be modified. It's different for subapps. that's why `req.originalUrl` exists

packages/cookie/src/index.ts
```diff
@@ -14,7 +14,7 @@ const fieldContentRegExp = /^[\u0009\u0020-\u007e\u0080-\u00ff]+$/
 function tryDecode(str: string, decode: (str: string) => string) {
   try {
     return decode(str)
-  } catch (e) {
+  } catch (_e) {
```
talentlessguy, line 17: can be replaced with `catch {`

tests/modules/proxy-addr.test.ts
```diff
@@ -517,15 +525,15 @@ describe('proxyaddr.all(req, trust?)', () => {
...
-      })
+      // it('should be required', () => {
+      //   try {
+      //     compile(null)
+      //   } catch (error) {
+      //     expect(error).toBeDefined()
+      //     return
+      //   }
+      //   assert.fail()
+      // })
```
talentlessguy, line 536: why comment a test out?

PR #420, opened 2023-10-26
fix: correctly update req.url on routes and mws

### LINKED ISSUE #417: Wildcards strip the first character from `req.url`
**Describe the bug**

Wildcards (`*`) don't correctly populate the URL on `/`-prefixed paths: they strip the first character.

**To Reproduce**

```js

const app = new App()

app.get('/foo/*',(req,res) => {
  console.log(req.url);
  // on GET /foo/bar
  // should show @/bar
  //   will show @/ar
  res.send(`@${req.url}`);
});

app.listen(3001);
```

**Expected behavior**

Should show `@/bar`

**Versions**

- `node`: 16
- `@tinyhttp/app`: 2.2.0

adhusson:
Closes #417

### DIFF
--- a/packages/app/src/app.ts
+++ b/packages/app/src/app.ts
@@ -358,14 +358,22 @@ export class App<Req extends Request = Request, Res extends Response = Response>
         else throw e
       }
 
+      // Warning: users should not use :wild as a pattern
+      let prefix = path
+      if (regex) {
+        for (const key of regex.keys) {
+          if (key === 'wild') {
+            prefix = prefix.replace('*', params.wild)
+          } else {
+            prefix = prefix.replace(`:${key}`, params[key])
+          }
+        }
+      }
+
       req.params = { ...req.params, ...params }
 
-      if (path.includes(':')) {
-        const first = Object.values(params)[0]
-        const url = req.url.slice(req.url.indexOf(first) + first?.length)
-        req.url = lead(url)
-      } else {
-        req.url = lead(req.url.substring(path.length))
+      if (mw.type === 'mw') {
+        req.url = lead(req.originalUrl.substring(prefix.length))
       }
 
       if (!req.path) req.path = getPathname(req.url)
--- a/tests/core/request.test.ts
+++ b/tests/core/request.test.ts
@@ -44,6 +44,43 @@ describe('Request properties', () => {
 
       await fetch('/abc/def').expect(200, '/def')
     })
+    it('should set the correct req.url on routes even in a subapp', async () => {
+      const makeApp = () =>
+        new App().get('/a1/b/*', (req, res) => res.send(req.url)).get('/a2/b/:pat', (req, res) => res.send(req.url))
+
+      const app = makeApp()
+      app.use('/s/:pat1/:pat2', makeApp())
+      const fetch = makeFetch(app.listen())
+      await fetch('/a1/b/c').expect(200, '/a1/b/c')
+      await fetch('/a2/b/c').expect(200, '/a2/b/c')
+      await fetch('/s/t/u/a1/b/c').expect(200, '/a1/b/c')
+      await fetch('/s/t/u/a2/b/c').expect(200, '/a2/b/c')
+    })
+    it('should set the correct req.url on middlewares even in a subapp', async () => {
+      const mw = (req, res, next) => {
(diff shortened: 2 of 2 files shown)


tests/core/request.test.ts
```diff
@@ -44,6 +44,43 @@ describe('Request properties', () => {
 
       await fetch('/abc/def').expect(200, '/def')
     })
+    it('should set the correct req.url on routes even in a subapp', async () => {
+      const makeApp = () =>
+        new App().get('/a1/b/*', (req, res) => res.send(req.url)).get('/a2/b/:pat', (req, res) => res.send(req.url))
+
+      const app = makeApp()
+      app.use('/s/:pat1/:pat2', makeApp())
```
aarontravass, line 52: Add a test case for when a subapp is mount on a wildcard route.

aarontravass:
@talentlessguy

packages/app/src/app.ts
```diff
@@ -358,14 +358,22 @@ export class App<Req extends Request = Request, Res extends Response = Response>
         else throw e
       }
 
+      // Warning: users should not use :wild as a pattern
+      let prefix = path
+      if (regex) {
+        for (const key of regex.keys) {
+          if (key === 'wild') {
```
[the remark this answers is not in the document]
adhusson: It is covered in `'should set the correct req.url on middlewares even in a subapp'`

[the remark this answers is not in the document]
adhusson: This is just a general warning, the issue already existed before this PR (it clashes with the `'wild'`` key used when there is a `*` in the pattern. 

There could be a separate PR that adds a test showing that this does indeed fail but it seems out of scope of the current PR.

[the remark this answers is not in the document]
aarontravass: alright.

tests/core/request.test.ts
adhusson: Done in 7c3896fc4ac60aab28a92701622117b65482a08a, but note that since `*` is greedy the only possible paths for those subapps is `/`

tests/core/request.test.ts
```diff
@@ -44,6 +44,43 @@ describe('Request properties', () => {
...
+    it('should set the correct req.url on middlewares even in a subapp', async () => {
+      const mw = (req, res, next) => {
+        req.urls ||= []
+        req.urls.push(req.url)
+        next()
+      }
+      const makeApp = () =>
+        new App()
+          .use('/a1/b', (req, res) => res.send(req.url))
+          .use('/a2/b', mw, mw, mw, (req, res) => res.send(`${req.urls.join(' ')}`))
```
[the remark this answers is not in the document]
adhusson: Done in 7c3896fc4ac60aab28a92701622117b65482a08a

tests/core/request.test.ts
```diff
@@ -44,6 +44,109 @@ describe('Request properties', () => {
...
+
+      await fetch('/s/t/a4/b/c/d').expect(200, {
+        url: '/',
+        params: { pat: 't', pat1: 'b', wild: 'c/d' }
+      })
+    })
+    it('should set the correct req.url on a subapp mounted on a wildcard route, for both route and mw', async () => {
+      const echo = (req, res) => res.send({ url: req.url, params: req.params })
+      // Only possible route on subapps below * is / since * is greedy
+      const subAppRoute = new App().get('/', echo)
```
aarontravass, line 137: in `.get('/', echo)` use something other than `/` since `/` is the default and it is difficult to know if the route is resolving correctly.

tests/core/request.test.ts
```diff
@@ -44,6 +44,109 @@ describe('Request properties', () => {
...
+      const subAppMw = new App().use('/', echo)
```
aarontravass, line 138: Same as above.

aarontravass (changes requested):

tests/core/request.test.ts
adhusson: It cannot work since we are below a `*`. The entire URL is matched and the only thing that can be matched in a subapp mounted under a wildcard is `/`.

talentlessguy:
safe to merge @adhusson ?

adhusson:
yes!

PR #434, opened 2024-07-27
Fix/subapps blocking middleware

### LINKED ISSUE #433: app.route blocks middleware mounted at a subpath
# Case 1
## To Reproduce
```ts

const app = new App()

app.route('/')
  .get((req, res) => void res.send(`this is rooty`))

app.use('/foo', (req, res) => void res.send(`this is not rooty`))

app.listen(3000)
```

## Result
- [http://localhost:3000](http://localhost:3000) yields `this is rooty`
- [http://localhost:3000/foo](http://localhost:3000/foo) yields `Not Found`

## Expected behavior
- [http://localhost:3000](http://localhost:3000) yields `this is rooty`
- [http://localhost:3000/foo](http://localhost:3000/foo) yields `this is not rooty`

# Case 2 
## To Reproduce
```ts

const app = new App()

app.use('/', (req, res) => void res.send(`this is rooty`))

app.use('/foo', (req, res) => void res.send(`this is not rooty`))

app.listen(3000)
```

## Result
- [http://localhost:3000](http://localhost:3000) yields `this is rooty`
- [http://localhost:3000/foo](http://localhost:3000/foo) yields `this is rooty`

## Expected behavior
- [http://localhost:3000](http://localhost:3000) yields `this is rooty`
- [http://localhost:3000/foo](http://localhost:3000/foo) yields `this is not rooty`

## Evironment

- `node`: 20.11.1
- `@tinyhttp/app`: 2.3.0

Lordfirespeed:
Resolves #433

### DIFF
--- a/packages/app/src/app.ts
+++ b/packages/app/src/app.ts
@@ -63,7 +63,7 @@ export class App<Req extends Request = Request, Res extends Response = Response>
   settings: AppSettings
   engines: Record<string, TemplateEngine> = {}
   applyExtensions: (req: Request, res: Response, next: NextFunction) => void
-  attach: (req: Req, res: Res) => void
+  attach: (req: Req, res: Res, next?: NextFunction) => void
   cache: Record<string, unknown>
 
   constructor(options: AppConstructor<Req, Res> = {}) {
@@ -79,7 +79,8 @@ export class App<Req extends Request = Request, Res extends Response = Response>
       ...options.settings
     }
     this.applyExtensions = options?.applyExtensions
-    this.attach = (req, res) => setImmediate(this.handler.bind(this, req, res, undefined), req, res)
+    const boundHandler = this.handler.bind(this)
+    this.attach = (req, res, next?: NextFunction) => setImmediate(boundHandler, req, res, next)
     this.cache = {}
   }
 
@@ -152,7 +153,7 @@ export class App<Req extends Request = Request, Res extends Response = Response>
 
   /**
    * Render a template
-   * @param file What to render
+   * @param name What to render
    * @param data data that is passed to a template
    * @param options Template engine options
    * @param cb Callback that consumes error and html
@@ -300,6 +301,7 @@ export class App<Req extends Request = Request, Res extends Response = Response>
    * Extends Req / Res objects, pushes 404 and 500 handlers, dispatches middleware
    * @param req Req object
    * @param res Res object
+   * @param next 'Next' function
    */
   handler<RenderOptions extends TemplateEngineOptions = TemplateEngineOptions>(
     req: Req,
@@ -387,17 +389,30 @@ export class App<Req extends Request = Request, Res extends Response = Response>
 
     let idx = 0
 
-    const loop = () => res.writableEnded || (idx < mw.length && handle(mw[idx++])(req, res, next))
+    const handleNext = (): void => void handle(mw[idx++])(req, res, next)
 
-    next = next || ((err) => (err ? this.onError(err, req, res) : loop()))
+    const parentNext = next
+    next = (err) => {
+      if (err != null) {
+        return this.onError(err, req, res)
+      }
 
-    loop()
+      if (res.writableEnded) return
+      if (idx >= mw.length) {
+        if (parentNext != null) parentNext()
+        return
+      }
+
+      handleNext()
+    }
+
+    handleNext()
   }
 
   /**
    * Creates HTTP server and dispatches middleware
    * @param port server listening port
-   * @param Server callback after server starts listening
+   * @param cb callback to be invoked after server starts listening
    * @param host server listening host
    */
   listen(port?: number, cb?: () => void, host?: string): Server {
--- a/packages/app/src/app.ts
+++ b/packages/app/src/app.ts
@@ -343,11 +343,13 @@ export class App<Req extends Request = Request, Res extends Response = Response>
       })
     }
 
-    mw.push({
-      handler: this.noMatchHandler,
-      type: 'mw',
-      path: '/'
-    })
+    if (this.parent == null) {
+      mw.push({
+        handler: this.noMatchHandler,
+        type: 'mw',
+        path: '/'
+      })
+    }
 
     const handle = (mw: Middleware) => async (req: Req, res: Res, next?: NextFunction) => {
       const { path, handler, regex } = mw
--- a/packages/app/src/app.ts
+++ b/packages/app/src/app.ts
@@ -391,7 +391,7 @@ export class App<Req extends Request = Request, Res extends Response = Response>
 
     let idx = 0
 
-    const handleNext = (): void => void handle(mw[idx++])(req, res, next)
+    const loop = (): void => void handle(mw[idx++])(req, res, next)
 
     const parentNext = next
     next = (err) => {
@@ -405,10 +405,10 @@ export class App<Req extends Request = Request, Res extends Response = Response>
         return
       }
 
-      handleNext()
+      loop()
     }
 
-    handleNext()
+    loop()
   }
 
   /**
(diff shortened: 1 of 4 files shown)


tests/core/app.test.ts
```diff
@@ -683,6 +685,23 @@ describe('Subapps', () => {
 
     await fetch('/').expect(200, 'Hello World!')
   })
+  it('sub-app mounted on root and subpath', async () => {
```
aarontravass, line 688: Can you rename this test to reflect the actual test?

packages/app/src/app.ts
```diff
@@ -387,17 +391,30 @@ export class App<Req extends Request = Request, Res extends Response = Response>
 
     let idx = 0
 
-    const loop = () => res.writableEnded || (idx < mw.length && handle(mw[idx++])(req, res, next))
+    const handleNext = (): void => void handle(mw[idx++])(req, res, next)
```
aarontravass, line 394: Type annotation can be removed.

aarontravass (changes requested):

aarontravass:
Can you add more verb specific (`app.route`, `app.all` etc) test cases?

tests/core/app.test.ts
Lordfirespeed: I don't like that name either - just trying my best to follow the convention that the similar tests follow.

If you give me the particular name you want me to use, sure. 

I don't fancy playing that guessing game presently 😅

packages/app/src/app.ts
Lordfirespeed: it *can*, but why be implicit when you can be explicit? 

Annotating `void` ensures that the signature is not accidentally changed in future.

Lordfirespeed:
> Can you add verb specific (`app.route`, `app.all` etc) test cases? 

`.route` is an alias for `.use(new App)`, so I don't see the point.

`.all`, `.get` etc are out of scope for this PR, as explained in #433

aarontravass:
> Yes, I'll add a test for one verb-specific method (`get`) and the verb catch-all method

Sounds good. Thanks.

tests/core/app.test.ts
aarontravass: Maybe something along the lines of `should search for routes in other subapps even if a subapp mounted on '/' is present`. What do you reckon?

aarontravass:
> `.route` is an alias for `.use(new App)`, so I don't see the point.

To be honest, I wouldn't mind adding it as a sanity check.

tests/core/app.test.ts
Lordfirespeed: Aye, I catch your drift. I might go for something like this:
```ts
describe("when a subapp is mounted on '/'", () => {
  it("should continue middleware execution when subapp middleware is exhausted", () => { ... })
  // other related tests
})
```

aarontravass: Great! Are there other test cases which you can group under `when a subapp is mounted on '/'`? It wouldn't make sense to have a test suite with only 1 test.

Lordfirespeed: > Great! Are there other test cases which you can group under `when a subapp is mounted on '/'`? It wouldn't make sense to have a test suite with only 1 test.

There are, so this is moot, but regardless - I disagree, using `describe` to provide context to a single testcase is a common practise. See Mocha's `context` ([example](https://mochajs.org/#bdd>))

Lordfirespeed:
My only remaining concern at this point is that it's possible I've significantly (detrimentally) affected framework benchmarks - I'm not sure how to test whether or not that is the case, though 😅

tests/core/app.test.ts
aarontravass: Cool.

talentlessguy:
@Lordfirespeed  you can use https://github.com/mcollina/autocannon but keep in mind that it's results are unreliable because network has a bunch of side effects

packages/app/src/app.ts
```diff
@@ -79,7 +79,8 @@ export class App<Req extends Request = Request, Res extends Response = Response>
       ...options.settings
     }
     this.applyExtensions = options?.applyExtensions
-    this.attach = (req, res) => setImmediate(this.handler.bind(this, req, res, undefined), req, res)
+    const boundHandler = this.handler.bind(this)
+    this.attach = (req, res, next?: NextFunction) => setImmediate(boundHandler, req, res, next)
```
talentlessguy, line 83: no need to move it to a separate variable cuz it's used only once

packages/app/src/app.ts
```diff
@@ -152,7 +153,7 @@ export class App<Req extends Request = Request, Res extends Response = Response>
 
   /**
    * Render a template
-   * @param file What to render
+   * @param name What to render
```
talentlessguy, line 156: seems like unrelated change

packages/app/src/app.ts
```diff
@@ -387,17 +391,30 @@ export class App<Req extends Request = Request, Res extends Response = Response>
 
     let idx = 0
 
-    const loop = () => res.writableEnded || (idx < mw.length && handle(mw[idx++])(req, res, next))
```
talentlessguy, line 390: a function rename is unnecessary

packages/app/src/app.ts
```diff
@@ -387,17 +391,30 @@ export class App<Req extends Request = Request, Res extends Response = Response>
...
+    }
 
-    loop()
+    handleNext()
   }
 
   /**
    * Creates HTTP server and dispatches middleware
    * @param port server listening port
-   * @param Server callback after server starts listening
```
talentlessguy, line 400: unrelated change

talentlessguy (changes requested):

packages/app/src/app.ts
Lordfirespeed: The difference is that `.bind` is invoked on every invocation when the variable is inlined - in theory, what I have done saves memory

packages/app/src/app.ts
Lordfirespeed: My IDE was giving me grief about the jsdoc parameter being missing. Moving this change (and the other jsdoc one above) to another PR would be a waste of my time.

packages/app/src/app.ts
Lordfirespeed: My IDE was giving me grief about the jsdoc parameter being missing.

packages/app/src/app.ts
Lordfirespeed: Happy to revert, but maybe we can agree on a more suitable name? Both 'loop' and 'handleNext' are not great at explaining what it does.

talentlessguy: I don't know about a more suitable name but renaming stuff isn't in the scope of this PR. The PR is fixing a bug.

packages/app/src/app.ts
talentlessguy: ok

packages/app/src/app.ts
Lordfirespeed: That logic is really flawed - in order to fix a bug, often you have to refactor things. Trying to keep names the same whilst changing behaviour means intent/behaviour will diverge from the variable/function names, reducing code quality.

You've expressed you aren't interested in trying to agree on a better name, so I'll revert to `loop`

talentlessguy: it's not flawed, the function name "loop" corresponds to middleware dispatch loop. That's why it's called as such. It was taken from Polka's source where it does the same thing and is called the same

Lordfirespeed: No, that's not what I mean. I see why it's called `loop` and I think that logic is sound (though I don't think the name is perfect).

The logic of 'this PR is a fix, so we can't rename anything' is what I am referring to.

Lordfirespeed:
Gave `autocannon` a go and I see what you mean - it's quite variable - that said, I can't see significant differences between `master` and this PR, so performance concerns can probably be put aside.

talentlessguy (approved):

PR #439, opened 2024-08-03
Feat/`req.port`

### LINKED ISSUE #437: More network extensions
When using a trusted reverse proxy, it can be tricky to identify the request port.

For example:
- user requests `https://foo.com:8901`
- reverse proxy server forwards request to `localhost:3000` and attaches `X-Forwarded-Host` header with value `foo.com:8901`
- `req.hostname` reads `foo.com` 
- identifying the request port is contrived (involves reading `X-Forwarded-Host` header **only when the proxy is trusted**) although the framework has already done the required processing

Proposal:
- add `request.port` network extension (self-explanatory, e.g. `8901`)
- add `request.host` network extension (combination of hostname and port, e.g. `foo.com:8901`)
- add `request.origin` network extension (combination of protocol, hostname and port, e.g.`https://foo.com:8901`
- add `request.href` network extension (combination of origin and path, e.g. `https://foo.com:8901/path` 

To match the [Node.js URL API](https://nodejs.org/api/url.html).

Lordfirespeed:
Resolves #437


packages/app/src/request.ts
```diff
@@ -32,17 +37,32 @@ export const getProtocol = (req: Request, trust: Trust): Protocol => {
   return index !== -1 ? header.substring(0, index).trim() : header.trim()
 }
 
+/**
+ * @deprecated
+ *
+ * @param req
+ * @param trust
```
talentlessguy, line 44: how is it deprecated? express `req.hostname` isn't

or you want to deprecate it because getHostname already obtains the host? Either way, hostname and host are different. hostname includes only the name, aka `localhost`, but host includes port too, aka `localhost:3000`

packages/app/src/request.ts
```diff
@@ -14,6 +14,11 @@ import type { URLParams } from '@tinyhttp/req'
 
 export { getURLParams } from '@tinyhttp/req'
 
+export type Host = {
```
aarontravass, line 17: You don't have to export this type.

packages/app/src/request.ts
```diff
@@ -32,17 +37,32 @@ export const getProtocol = (req: Request, trust: Trust): Protocol => {
...
+ *
+ * @param req
+ * @param trust
+ */
 export const getHostname = (req: Request, trust: Trust): string | undefined => {
+  const { hostname } = getHost(req, trust)
+  return hostname
+}
+
+export const getHost = (req: Request, trust: Trust): Host | undefined => {
```
aarontravass, line 51: `export const getHost = (req: Request, trust: Trust): { hostname: string, port?: number } | undefined => {`

This way, you won't have to define a new type.

aarontravass (changes requested):

tests/core/request.test.ts
```diff
@@ -237,6 +237,30 @@ describe('Request properties', () => {
 
       await fetch('/').expect(200, 'subdomains: ')
     })
+    it('assigns req.hostname', async () => {
```
aarontravass, line 240: Change the test name.

tests/core/request.test.ts
```diff
@@ -237,6 +237,30 @@ describe('Request properties', () => {
...
+          res.send(`hostname: ${req.hostname}`)
+        },
+        '/',
+        'GET',
+        options
+      )
+
+      await fetch('/', { headers: { Host: 'foo.bar:8080' } }).expect(200, 'hostname: foo.bar')
+    })
+    it('assigns req.port', async () => {
```
aarontravass, line 252: Same as above.

aarontravass (changes requested):

aarontravass:
Can you add tests for the following cases:
1. `host` header is absent
2. server is behind a proxy

packages/app/src/request.ts
Lordfirespeed: I marked it as deprecated because it's superceded by the `getHost` method added below - `getHost` returns an object with `hostname` and `port` properties, so the hostname can be retrieved by object destructuring
```ts
const { hostname } = getHost(...)
```

packages/app/src/request.ts
Lordfirespeed: what's wrong with defining the type explicitly? inlining the type only seems to obfuscate and increase the line-length

tests/core/request.test.ts
Lordfirespeed: To ...? have you looked at the test names in context

tests/core/request.test.ts
Lordfirespeed: Same as [this](https://github.com/tinyhttp/tinyhttp/pull/439/files#r1702860768)

tests/core/request.test.ts
aarontravass: `should assign hostname to req.hostname`

tests/core/request.test.ts
aarontravass: `should assign the correct port to req.port by extracting it from the Host header`

packages/app/src/request.ts
aarontravass: `type Host` is not reused.

Lordfirespeed:
> Can you add tests for the following cases:
> 
>     1. `host` header is absent

If the `Host` header is absent the request should always be aborted (400 Bad Request) - a test for that seems not relevant here?

>     2. server is behind a proxy

That case is handled entirely by the `proxy-addr` module, for which there are tests for that

packages/app/src/request.ts
aarontravass: `getHostName` is an internal method and not supposed to be used by the end user, hence marking it as deprecated is rather pointless.

aarontravass:
> If the `Host` header is absent the request should always be aborted (400 Bad Request) - a test for that seems not relevant here?

How about tests checking if the host header is malformed? For example `Host: localhost:hello` or `Host: localhost`

tests/core/request.test.ts
aarontravass: @Lordfirespeed I hope the test name is pellucid. It tells you what test is being performed.

Lordfirespeed:
> How about tests checking if the host header is malformed? For example `Host: localhost:hello` or `Host: localhost`

I agree those should be added to #92.

packages/app/src/request.ts
Lordfirespeed: Many functions are only used once, but that is no reason to inline them

packages/app/src/request.ts
Lordfirespeed: OK, I'll remove it :)

aarontravass:
> > How about tests checking if the host header is malformed? For example `Host: localhost:hello` or `Host: localhost`
> 
> I agree those should be added to #92.

Your code introduces this check:

https://github.com/Lordfirespeed/tinyhttp/blob/feat/req-port/packages/app/src/request.ts#L64

and it would be better to write a case which covers this branch.

> Edit: Wouldn't `Host: localhost` be perfectly valid? assuming the protocol is `http`

Shouldn't it return a port? https://httpwg.org/specs/rfc9110.html#field.host

Lordfirespeed:
> Your code introduces this check [...] and it would be better to write a case which covers this branch.

That's a good point, well made

> Shouldn't it return a port? https://httpwg.org/specs/rfc9110.html#field.host

```
Host = uri-host [ ":" port ]
```
The `[]` indicate the `:` and port are optional - can you clarify what you mean by 'it'? are you referring to the `getHost` function?

aarontravass:
> can you clarify what you mean by 'it'?

`req.port`. What happens if the host header only contains `google.com` and the user logs `req.port`? 

Also, port can be obtained from `:authority` header. I would encourage you to check the port in `:authority` as a fallback.

aarontravass:
> Also, port can be obtained from `:authority` header. I would encourage you to check the port in `:authority` as a fallback.

For context, see:

> In HTTP/2 [[HTTP/2]](https://httpwg.org/specs/rfc9110.html#HTTP2) and HTTP/3 [[HTTP/3]](https://httpwg.org/specs/rfc9110.html#HTTP3), the Host header field is, in some cases, supplanted by the ":authority" pseudo-header field of a request's control data.

https://httpwg.org/specs/rfc9110.html#field.host

aarontravass:
Here's some pseudo code to help you

```ts
const portFromHostHeader // extract from host
if(portFromHostHeader) return portFromHostHeader
const portFromAuthorityHeader // extract from :authority header
if(portFromAuthorityHeader) return portFromAuthorityHeader 
return null
```

Lordfirespeed:
> What happens if the host header only contains `google.com` and the user logs `req.port`?

`req.port` will be `undefined`, which is fine because either
- the consumer can determine the port using `req.protocol`
- the consumer does not need to determine the port because they can assume the client will continue to (correctly) default the port on subsequent requests

> Also, port can be obtained from `:authority` header. I would encourage you to check the port in `:authority` as a fallback.

Good point - I think out of scope for this PR, though, because the `getHostname` method was not previously respecting that header either

packages/app/src/request.ts
```diff
@@ -32,17 +37,21 @@ export const getProtocol = (req: Request, trust: Trust): Protocol => {
...
+  if (!host) return undefined
 
   // IPv6 literal support
   const index = host.indexOf(':', host[0] === '[' ? host.indexOf(']') + 1 : 0)
+  if (index === -1) return { hostname: host }
 
-  return index !== -1 ? host.substring(0, index) : host
+  const hostname = host.substring(0, index)
+  const port = Number(host.substring(index + 1))
+  if (Number.isNaN(port)) throw new TypeError('Port number is NaN, therefore Host is malformed')
```
aarontravass, line 53: Just realized something: it's better to avoid throwing an error cause if during request lifecycle, an invalid host header is sent, the end user's app will crash. Instead, return `null`.

aarontravass (changes requested):

packages/app/src/request.ts
Lordfirespeed: I don't think that's true, I'm pretty confident the error will be caught by the top-level error handler, leading to a `500 Internal Server Error` response

aarontravass: Can you add a test case for concreteness?

Lordfirespeed:
I have set the `:authority` headers to be skipped because the Fetch API outright forbids assigning the `:authority` header manually, and I'm not sure how to work around that.

aarontravass:
> I have set the `:authority` header tests to be skipped because the node.js Fetch API outright forbids assigning the `:authority` header manually, and I'm not sure how to work around that.

Maybe try `new Headers().set....`

Lordfirespeed:
> Maybe try `new Headers().set....`

Already did, no dice

aarontravass:
> > Maybe try `new Headers().set....`
> 
> Already did, no dice

Can you post the error?

aarontravass:
> > > Maybe try `new Headers().set....`
> > 
> > 
> > Already did, no dice
> 
> Can you post the error?

Nevermind, I could reproduce it.

aarontravass:
> > Maybe try `new Headers().set....`
> 
> Already did, no dice

Maybe try mocking fetch? I'm out of ideas.

Lordfirespeed:
> Maybe try mocking fetch? I'm out of ideas.

That or actually using a HTTP2 server 😅  
I tried `undici.request` but also no joy there

Lordfirespeed:
@aarontravass take a look at the workaround, let me know if it's sufficient :sob:

tests/fixtures/server.crt
```diff
@@ -0,0 +1,19 @@
+-----BEGIN CERTIFICATE-----
```
aarontravass, line 1: I'm wondering if there's a way to generate the key/cert on the fly.

tests/core/request.test.ts
```diff
@@ -261,6 +293,50 @@ describe('Request properties', () => {
...
+          connect: {
+            rejectUnauthorized: false
+          },
+          allowH2: true
+        })
+      )
+
+      const { server } = InitSecureAppAndTest(
+        (req, res) => {
+          //expect(req.get('host')).toBeUndefined()
```
aarontravass, line 313: nitpick: remove the comment.

aarontravass:
> > Maybe try mocking fetch? I'm out of ideas.
> 
> That or actually using a HTTP2 server 😅 I tried `undici.request` but also no joy there

Was thinking of a third party module but refrained from mentioning cause it would've added a dev dep. It's fine I guess.

Lordfirespeed:
`undici` is actually installed anyway, it's what backs the native node fetch API - but I take your point

tests/core/request.test.ts
Lordfirespeed: This isn't supposed to be commented out, good catch

tests/fixtures/server.crt
Lordfirespeed: Yes, but it'd involve calling `openssl` from the command line and would increase the test suite runtime

aarontravass: Alrighty.

aarontravass:
@talentlessguy

talentlessguy:
for some reason the coverage dropped, could you please ensure that all the new code is covered in tests?

aarontravass:
@Lordfirespeed CI is failing

Lordfirespeed:
> @Lordfirespeed CI is failing

it's been 10 minutes :sob:

Lordfirespeed:
Yeah, there's not much I can do about that. `fetch` API makes it very tricky to test that kind of case.

aarontravass:
> @talentlessguy Coverage dropped because of lines handling a difficult-to-test case - when the `:authority` header disagrees with the `host` header

https://github.com/nghttp2/nghttp2/issues/317#issuecomment-128023377

aarontravass:
> @talentlessguy Coverage dropped because of lines handling a difficult-to-test case - when the `:authority` header disagrees with the `host` header

Controversial take: I feel like the server should not validate request headers and instead, the burden should be passed over to the end user.

Lordfirespeed:
> I feel like the server should not validate request headers and instead, the burden should be passed over to the end user.

Then the burden of the network extensions should also be passed to the end user, and this whole PR is moot. 

We can't (definitively) figure out the request hostname without validating the request, because the hostname is not well-defined for requests that don't adhere to the HTTP spec.

talentlessguy:
you can mock the part of the test for `:authority` headers using `vi.fn`, I don't have a problem with that.

aarontravass:
> Then the burden of the network extensions should also be passed to the end user, and this whole PR is moot.

That is not what I meant. Even if `req.port` is undefined or null, it should be the end user who gets decide whether the server should respond with status code `400` or something else. The server should parse the request object and try its best to populate those fields but if a particular field is absent, the burden shouldn't lie on the server for request validation. Parsing != Validation.

For example, if the host header contained the following `google.com:100000000`, all that the server needs to do is, populate the `hostname` and `port` options with `google.com` and `100000000` respectively, even though the port is clearly invalid. 

It's just my personal opinion. 

https://github.com/expressjs/express/blob/5.0/lib/request.js#L203

Lordfirespeed:
> it should be the end user who gets decide whether the server should respond with status code 400 or something else

Which consumers are able to do by implementing the `onError` function.

Parsing is not equal to validation, yes. But parsing relies on validation. 
You can't run a JavaScript parser on C code. It fails because the input is invalid.

aarontravass:
> > it should be the end user who gets decide whether the server should respond with status code 400 or something else
> 
> Which consumers are able to do by implementing the `onError` function.
> 
> Parsing is not equal to validation, yes. But parsing relies on validation. You can't run a JavaScript parser on C code. It fails because the input is invalid.

I agree. 

Maybe I'm dumb. Educate me on the purpose of that check.

Lordfirespeed:
> Educate me on the purpose of that check.

Certainly! That check exists because of a clause in the [HTTP/2 spec](https://datatracker.ietf.org/doc/html/rfc9113#section-8.3.1)

`
Clients MUST NOT generate a request with a Host header field that differs from the ":authority" pseudo-header field. A server SHOULD treat a request as malformed if it contains a Host header field that identifies an entity that differs from the entity in the ":authority" pseudo-header field.
`

i.e., if the `Host` header is present and disagrees with the `:authority` header, a HTTP/2 compliant server should consider the request invalid, and respond with status `400`.

aarontravass:
> `Clients MUST NOT generate a request with a Host header field that differs from the ":authority" pseudo-header field. A server SHOULD treat a request as malformed if it contains a Host header field that identifies an entity that differs from the entity in the ":authority" pseudo-header field.`

Damn. I'm sorry for not reading more on this.

vite.config.ts
talentlessguy, line 1: this change has 0 benefit and is unrelated to the pull request, please revert

talentlessguy (changes requested):

.github/workflows/main.yml
```diff
@@ -54,9 +54,6 @@ jobs:
       - name: install dependencies
         run: pnpm install
 
-      - name: build packages
```
talentlessguy, line 57: why did you remove a build step from CI?

Lordfirespeed: I assumed you only had it there to enable `pnpm test` to run without problems - because of the path aliases I added to `vitest.config.ts`, the build step is not required for tests anymore.

talentlessguy: no, it builds in CI to ensure that building packages doesn't fail

vite.config.ts
Lordfirespeed: No, it's very relevant. In order to mock `getRequestHeader` I need to ensure I import the same `@tinyhttp/req` from the test file as is transitively imported by `@tinyhttp/app`.

Without these path aliases, `@tinyhttp/app` imports `getRequestHeader` from `packages/req/dist`, so the mocking won't work right.

.github/workflows/main.yml
Lordfirespeed: I'll revert, then

vite.config.ts
talentlessguy: root vite config is not used anywhere in the build step. each package has an individual vite config.

instead of mocking the whole req package, you can extend an actual req package with your custom mock

docs: https://vitest.dev/api/vi.html#vi-mock

aarontravass: > Without these path aliases, `@tinyhttp/app` imports `getRequestHeader` from `packages/req/dist`, so the mocking won't work right.

vitest should automatically transpile them every time you run `pnpm test`

Lordfirespeed:
FYI: Coverage has dropped due to updating the coverage reporter

vite.config.ts
Lordfirespeed: Agreed r.e. `vitest` vs `vite` config, but since it's only used for `vitest`, it makes more sense for it to be `vitest.config.ts`.

You're wrong, you have to mock the whole req package, since it is imported as `@tinyhttp/req` from the `app` package.
The example you've linked is an example of mocking the whole package.

Lordfirespeed: Without path aliases: 
image

With path aliases:
image

Lordfirespeed: > vitest should automatically transpile them every time you run `pnpm test`?

@aarontravass Read my messages again. You are correct, but you've misunderstood the issue i'm referring to

If we mock the module `packages/req/src/index`, but `@tinyhttp/app` imports `packages/req/dist/index`, `@tinyhttp/app` will be importing the **un-mocked** module, which leads the tests to fail. 

The path aliases force `@tinyhttp/app` to import `package/req/src/index`, etc.

test_helpers/initAppAndTest.ts
```diff
@@ -1,4 +1,7 @@
+import fs from 'node:fs/promises'
 import type { Server } from 'node:http'
+import { type Http2SecureServer, createSecureServer } from 'node:http2'
```
talentlessguy, line 3: these imports aren't used anywhere

Lordfirespeed: Yes, good catch

packages/app/src/request.ts
```diff
@@ -32,17 +37,64 @@ export const getProtocol = (req: Request, trust: Trust): Protocol => {
   return index !== -1 ? header.substring(0, index).trim() : header.trim()
 }
 
-export const getHostname = (req: Request, trust: Trust): string | undefined => {
-  let host: string = req.get('X-Forwarded-Host') as string
+const normalizeHostString = (host: string): string => decodeURIComponent(host).toLowerCase().normalize()
```
talentlessguy, line 40: why does host have to be uri decoded?

packages/app/src/request.ts
```diff
@@ -32,17 +37,64 @@ export const getProtocol = (req: Request, trust: Trust): Protocol => {
...
+
+const getAuthorityHeaderHostString = (req: Request): string | undefined => {
+  const authority = req.get(':authority')
+  if (Array.isArray(authority)) return undefined
+  if (!authority) return undefined
```
talentlessguy, line 45: can be simplified with 
```suggestion
  if (Array.isArray(authority) || !authority) return undefined
```

packages/app/src/request.ts
```diff
@@ -32,17 +37,64 @@ export const getProtocol = (req: Request, trust: Trust): Protocol => {
...
+  const forwardedHost = req.get('x-forwarded-host')
+  if (Array.isArray(forwardedHost)) return undefined
+  if (!forwardedHost) return undefined
+
+  return normalizeHostString(forwardedHost)
+}
+
+const getDefaultHeaderHostString = (req: Request): string | undefined => {
+  const host = req.get('host')
+  if (!host) return undefined
```
talentlessguy, line 62: can be simplified with:

```suggestion
  if (!host || host.indexOf(',') !== -1) return undefined
```

packages/app/src/request.ts
Lordfirespeed: [RFC 3986, 3.2.2. Host](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2) defines `host` syntactically
```
host = IP-literal / IPv4address / reg-name
...
reg-name    = *( unreserved / pct-encoded / sub-delims )
```
and Appendix A defines `pct-encoded`
```
pct-encoded   = "%" HEXDIG HEXDIG
```
i.e. standard URI encoding is valid for the `Host` header

talentlessguy: I see

talentlessguy:
sorry for the wait, I first want to fix strictness of TS and setup GHA for publish workflow, after that I'll re-review the PR

Lordfirespeed:
> first want to fix strictness of TS

For some solutions, feel free to look at the `fix/strict-type-checking` branch on my fork of this repo :)

talentlessguy:
I've bumped Vitest and it's c8 plugin, could you please git pull and copy paste the lockfile?

vitest.config.ts
```diff
@@ -0,0 +1,40 @@
...
+      '@tinyhttp/ip-filter': relative('packages/ip-filter/src'),
+      '@tinyhttp/jsonp': relative('packages/jsonp/src'),
+      '@tinyhttp/proxy-addr': relative('packages/proxy-addr/src'),
+      '@tinyhttp/rate-limit': relative('packages/rate-limit/src'),
+      '@tinyhttp/req': relative('packages/req/src'),
+      '@tinyhttp/res': relative('packages/res/src'),
+      '@tinyhttp/router': relative('packages/router/src'),
+      '@tinyhttp/send': relative('packages/send/src'),
+      '@tinyhttp/type-is': relative('packages/type-is/src'),
+      '@tinyhttp/url': relative('packages/url/src')
```
talentlessguy, line 37: this can be replaced with a loop I think. No need to hardcode aliases

talentlessguy (changes requested):

talentlessguy:
actually nvm I'll fix up the PR rn

talentlessguy (approved):

Issue #440, opened 2024-09-05
Support for url forwarding


aweigor:
Making changes to the `req.url` from inside of middleware does not affect the execution order.

Suppose such an example:
```
const app = new App()
const rewrite = (req, res, next) => {
  req.url = '/b'
  next()
}

app.use(rewrite).use('/a', (req, res) => {
  res.send({
    url: req.originalUrl
  })
}

app.listen(3000)
```

Actual output for `/a` is `{ url: '/b }`, while expected output is `{ url: '/a }` since the middleware was binded to the route `/a`.

It would be great if user could be redirected to the corresponding route by making changes in `req.url`.

aarontravass:
Hello @aweigor, thank you for raising this issue. When I ran the above code in express, it gave me 404 as the output
```html
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot GET /a</pre>
</body>
</html>
```

Since we are replicating the behavior of express, it would better to return 404 instead of a 200 response

aarontravass:
Also, the output I got for your code was
```json
{
  "url": "/a"
}
```

Maybe your package version is outdated?

aweigor:
Hello @aarontravass! Yes, if we'll define mw on route `/b` in express, the request will fall into it, otherwise 404 will be expected. For example:

```
const app = new express();
const rewrite = (req, res, next) => {
	req.url = "/b";
	next();
};

app.use(rewrite).use("/a", (req, res) => {
	res.send('Hello from a');
});

app.use(rewrite).use("/b", (req, res) => {
	res.send('Hello from b');
});

app.listen(3000);
``` 

You were right about my test package, now th responses with `\a`. Anyway, forwarding would be more preferable, i think.

aarontravass:
> forwarding would be more preferable, i think

I'm not sure. tinyhttp is supposed to be a drop in replacement and if an existing user is reliant on express's feature of returning 404 and now after replacing it with tinyhttp, they see a different behavior, it might be cause for concern.

aweigor:
Agree, that is exactly why we need forwarding. I suppose returning 404 is not quite the feature. It returns 404 because rewriting req.url to `\b` forwards to undefined route `/b` in the topic example. Please check express behavior with the same test but route `/a` replaced by `/b`.

```
const app = new App()
const rewrite = (req, res, next) => {
  req.url = '/b'
  next()
}

[...]

PR #441, opened 2024-09-06
Support for url forwarding (#440)

aweigor:


tests/core/app.test.ts
```diff
@@ -1315,4 +1315,17 @@ describe('App settings', () => {
...
+      req.url = '/a/b'
+      next()
+    }
+    const echo = (req, res) => res.send({ url: req.url })
+    const app = new App().use('/', rewrite).use('/a', echo)
+    const fetch = makeFetch(app.listen())
+    await fetch('/').expect(200, {
+      url: '/b'
+    })
+  })
```
aarontravass, line 1330: Can you add a 404 test as well?

aarontravass (changes requested):

tests/core/app.test.ts
aweigor: Added some tests.

aweigor:
If a new route is discovered, req.params will be reset to avoid confusion.

packages/app/src/app.ts
```diff
@@ -380,18 +355,54 @@ export class App<Req extends Request = Request, Res extends Response = Response>
...
 
-      if (!req.path) req.path = getPathname(req.url)
+      if (!req.path) req.path = pathname
 
       if (this.settings?.enableReqRoute) req.route = mw
 
       await applyHandler<Req, Res>(handler as unknown as Handler<Req, Res>)(req, res, next)
     }
 
+    const matchFilter = function (this: Middleware[], x: Middleware) {
```
talentlessguy, line 368: please change to arrow function

packages/app/src/app.ts
```diff
@@ -380,18 +355,54 @@ export class App<Req extends Request = Request, Res extends Response = Response>
...
+          type: 'mw',
+          handler: (req, res, next) => {
+            if (req.method === 'HEAD') {
+              res.statusCode = 204
+              return res.end('')
+            }
+            next()
+          },
+          path: '/'
+        })
```
talentlessguy, line 395:
```suggestion
        mw = [...matched, {
          type: 'mw',
          handler: (req, res, next) => {
            if (req.method === 'HEAD') {
              res.statusCode = 204
              return res.end('')
            }
            next()
          },
          path: '/'
        }]
```

talentlessguy (changes requested):

packages/app/src/app.ts
```diff
@@ -365,15 +365,15 @@ export class App<Req extends Request = Request, Res extends Response = Response>
...
+    const matchFilter = (x: Middleware) =>
       (req.method === 'HEAD' || (x.method ? x.method === req.method : true)) && !mw.includes(x)
 
     let idx = 0
 
     const loop = () => {
       req.originalUrl = req.baseUrl + req.url
       const pathname = getPathname(req.url)
-      const matched = this.#find(pathname).filter(matchFilter(mw))
+      const matched = this.#find(pathname).filter(matchFilter)
```
talentlessguy, line 376: if this is only used once, it can be inlined

aweigor: Agree, done

aweigor:
@talentlessguy here’s a little problem with typing, I could fix it if you want.

talentlessguy:
sure

talentlessguy (approved):

PR #445, opened 2024-09-21
feat: App interface

aweigor:
Added application interface for correct typing of req/res parameters of subapp callbacks.

### DIFF
--- a/packages/app/src/app.ts
+++ b/packages/app/src/app.ts
@@ -35,6 +35,7 @@ const applyHandler =
     }
   }
 
+
 /**
  * `App` class - the starting point of tinyhttp app.
  *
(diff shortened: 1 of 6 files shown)


packages/app/src/app.ts
```diff
@@ -87,64 +91,32 @@ export class App<Req extends Request = Request, Res extends Response = Response>
     this.cache = {}
   }
 
-  /**
-   * Set app setting
-   * @param setting setting name
-   * @param value setting value
-   */
   set<K extends keyof AppSettings>(setting: K, value: AppSettings[K]): this {
```
talentlessguy, line 94: i suppose jsdoc comments get inherited from the interface, is that correct?

aweigor: Definitely. We could take the description of the class (line 38)  into the interface too, but I don’t know if it would be ok..

aweigor: Jsdoc for set added to interface, thanks

talentlessguy:
this is not a fix but a feat, so it should have a minor bump

could you please run `pnpm chgset` and choose `@/app` for the minor bump?

aweigor:
got updated `@/app` and `@/jsonp` .json, both commited.

talentlessguy (approved):

PR #466, opened 2025-04-28
fix(content-disposition): allow empty filename without throwing error

ItayLeon-Max:
Description:
This pull request fixes an issue in the @tinyhttp/content-disposition package where parsing a Content-Disposition header with an empty filename would incorrectly throw an invalid extended field value error.
Now, if the filename* parameter value is empty, it will be properly handled and parsed as an empty string ('') instead of causing an exception.
This behavior aligns with the expectation that an empty filename is valid per the relevant RFC standards.

Changes:
	•	Updated parse function to treat empty extended parameters correctly.
	•	Adjusted decodefield to return an empty string when input is empty.
	•	Added safeguard for handling empty filename* values without throwing.

Impact:
This change ensures better compliance with the Content-Disposition specification and prevents runtime errors when dealing with empty filenames.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,5476 @@
+{
+  "name": "@tinyhttp/monorepo",
+  "version": "0.0.1",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@tinyhttp/monorepo",
+      "version": "0.0.1",
+      "license": "MIT",
+      "devDependencies": {
+        "@biomejs/biome": "1.9.4",
+        "@changesets/cli": "2.28.1",
+        "@commitlint/cli": "19.8.0",
+        "@commitlint/config-conventional": "19.8.0",
+        "@types/node": "22.10.7",
+        "@vitest/coverage-v8": "^3.1.1",
+        "@vitest/ui": "3.1.1",
+        "c8": "^10.1.3",
+        "eta": "2.2.0",
+        "header-range-parser": "^1.1.3",
+        "husky": "9.1.7",
+        "regexparam": "3.0.0",
+        "supertest-fetch": "1.5.0",
+        "typescript": "~5.7.3",
+        "vite": "^6.2.6",
+        "vitest": "3.1.1"
+      },
+      "engines": {
+        "node": ">=14.21.3",
+        "pnpm": ">=8"
+      }
+    },
+    "node_modules/@ampproject/remapping": {
+      "version": "2.3.0",
+      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
+      "integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@jridgewell/gen-mapping": "^0.3.5",
+        "@jridgewell/trace-mapping": "^0.3.24"
+      },
+      "engines": {
+        "node": ">=6.0.0"
+      }
+    },
+    "node_modules/@babel/code-frame": {
+      "version": "7.26.2",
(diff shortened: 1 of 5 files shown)


package-lock.json
talentlessguy, line 1: the project uses pnpm, not npm

packages/content-disposition/src/index.ts
```diff
@@ -82,6 +82,8 @@ function format({
 function createParams(filename?: string, fallback?: string | boolean) {
   if (filename === undefined) return {}
 
+  if (filename === '') return { filename: '' }
```
talentlessguy, line 85:
```suggestion
  if (filename === '') return { filename }
```

talentlessguy (changes requested):

package-lock.json
ItayLeon-Max: Thanks for pointing it out!
I’ve already removed the unnecessary package-lock.json file to match the project’s use of pnpm.

packages/content-disposition/src/index.ts
ItayLeon-Max: Thanks for the suggestion!
I’ve committed the suggested change – it indeed makes the code cleaner and avoids unnecessary duplication. 🙏🏻

packages/req/package.json
```diff
@@ -22,8 +22,10 @@
   },
   "dependencies": {
     "@tinyhttp/accepts": "workspace:*",
+    "@tinyhttp/req": "workspace:^",
     "@tinyhttp/type-is": "workspace:*",
     "@tinyhttp/url": "workspace:*",
-    "header-range-parser": "^1.1.3"
+    "header-range-parser": "^1.1.3",
+    "type": "^2.7.3"
```
talentlessguy, line 29: what's this package?

ItayLeon-Max: Sure! Here’s some context regarding the use of header-range-parser:

The header-range-parser package is used in the req module to parse the Range header from incoming requests. It supports partial content handling, such as file streaming. Since it’s imported directly in the module (used in getRangeFromHeader), it’s defined as a direct dependency.

Example use case:

const range = parseRange(fileSize, req.headers.range)
// -> [{ start: 0, end: 499 }]

This allows serving partial responses like:

HTTP/1.1 206 Partial Content
Content-Range: bytes 0-499/1000

talentlessguy: no i mean the 'type" package

ItayLeon-Max: Thanks for the clarification!

Looks like `"type"` was unintentionally added. It's not required by the module, so I'll remove it in a follow-up commit.

ItayLeon-Max:
Hope you’re doing well!
Just wanted to ask if everything looks good now and if we can go ahead and merge the pull request.
Please let me know if there’s anything else you’d like me to adjust or review.

Thanks again for your guidance!

pnpm-lock.yaml
```diff
@@ -159,6 +162,9 @@ importers:
       header-range-parser:
         specifier: ^1.1.3
         version: 1.1.3
+      type:
+        specifier: ^2.7.3
+        version: 2.7.3
```
talentlessguy, line 167: this package is still in the lockfile

talentlessguy (changes requested):

pnpm-lock.yaml
ItayLeon-Max: Thanks for pointing that out! I’ve just removed the leftover reference from the lockfile and pushed the updated version. Should be all clean now.

ItayLeon-Max: When you have a moment, could you please review everything once again?
If all looks good, I think we’re ready to proceed with merging the pull request! 🚀

Thanks a lot for your time and help!

ItayLeon-Max: Hey good morning, I pushed a new commit that removes all traces of the type package, including from the lockfile.
Let me know if there’s anything else you’d like me to update — otherwise, we should be good to proceed with the merge. Thanks!

packages/req/package.json
```diff
@@ -22,6 +22,7 @@
   },
   "dependencies": {
     "@tinyhttp/accepts": "workspace:*",
+    "@tinyhttp/req": "workspace:^",
```
talentlessguy, line 25: why do you add the package to itself?

talentlessguy (changes requested):

PR #494, opened 2026-03-27
fix: add TypeScript overload signatures for accepts methods

### LINKED ISSUE #396: Add missing `accepts` TypeScript signatures
Currently methods like `req.accepts` and others only have one signature defined in the `Request` interface. We should support all possible variations like it is in Express.

claygeo:
## Problem

The `Request` interface defines `accepts`, `acceptsEncodings`, `acceptsCharsets`, and `acceptsLanguages` with a single signature:

```typescript
accepts: (...types: string[]) => string | boolean | string[]
```

This means TypeScript can't narrow the return type based on usage. Users must cast or type-guard every call:

```typescript
// Current: always returns string | boolean | string[]
const type = req.accepts('json', 'html'); // string | boolean | string[]
// User has to cast: (type as string)

const all = req.accepts(); // string | boolean | string[] (should be string[])
```

The underlying `Accepts` class already handles all call patterns correctly and returns different types based on input.

## Solution

Add TypeScript overload signatures that narrow the return type:

```typescript
accepts: {
  (): string[]                           // no args = all accepted types
  (type: string): string | false         // single string = best match or false
  (types: string[]): string | false      // array = best match or false
  (...types: string[]): string | false   // variadic = best match or false
}
```

Also narrows `AcceptsReturns` from `boolean` to the literal `false` type, since these methods return `false` (not `true`) when no match is found. This matches Express's type definitions.

**No runtime changes.** The implementation already supports all call patterns. This is purely a type-narrowing improvement.

## Changes

- `packages/app/src/request.ts` — Added overload signatures for all four accepts methods, narrowed `AcceptsReturns` to use `false` instead of `boolean`

Closes #396

### DIFF
--- a/packages/app/src/extend.ts
+++ b/packages/app/src/extend.ts
@@ -78,10 +78,10 @@ export const extendMiddleware = <EngineOptions extends TemplateEngineOptions = T
       if (!acceptsInstance) acceptsInstance = new Accepts(req)
       return acceptsInstance
     }
-    req.accepts = (...types) => getAcceptsInstance().types(types)
-    req.acceptsCharsets = (...charsets) => getAcceptsInstance().charsets(charsets)
-    req.acceptsEncodings = (...encodings) => getAcceptsInstance().encodings(encodings)
-    req.acceptsLanguages = (...languages) => getAcceptsInstance().languages(languages)
+    req.accepts = ((...types: string[]) => getAcceptsInstance().types(types)) as Request['accepts']
+    req.acceptsCharsets = ((...charsets: string[]) => getAcceptsInstance().charsets(charsets)) as Request['acceptsCharsets']
+    req.acceptsEncodings = ((...encodings: string[]) => getAcceptsInstance().encodings(encodings)) as Request['acceptsEncodings']
+    req.acceptsLanguages = ((...languages: string[]) => getAcceptsInstance().languages(languages)) as Request['acceptsLanguages']
 
     req.xhr = checkIfXMLHttpRequest(req)
 
--- a/packages/app/src/extend.ts
+++ b/packages/app/src/extend.ts
@@ -79,9 +79,12 @@ export const extendMiddleware = <EngineOptions extends TemplateEngineOptions = T
       return acceptsInstance
     }
     req.accepts = ((...types: string[]) => getAcceptsInstance().types(types)) as Request['accepts']
-    req.acceptsCharsets = ((...charsets: string[]) => getAcceptsInstance().charsets(charsets)) as Request['acceptsCharsets']
-    req.acceptsEncodings = ((...encodings: string[]) => getAcceptsInstance().encodings(encodings)) as Request['acceptsEncodings']
-    req.acceptsLanguages = ((...languages: string[]) => getAcceptsInstance().languages(languages)) as Request['acceptsLanguages']
+    req.acceptsCharsets = ((...charsets: string[]) =>
+      getAcceptsInstance().charsets(charsets)) as Request['acceptsCharsets']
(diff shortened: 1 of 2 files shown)


claygeo:
Friendly bump on this one.

Rebased onto the latest `master` (drift cleared), CI is green and the branch is mergeable. While here I also did a small cleanup: removed the now-unused `AcceptsReturns` alias and wrapped the `acceptsCharsets/Encodings/Languages` cast lines to match biome's 120-char `lineWidth`.

Still closes #396 (open). The change is type-only with no runtime change. The existing test suite passes and the overloads narrow correctly: no-arg returns `string[]`, with-arg returns `string | false`, matching Express's typings.

@v1rtl when you have a moment, would you be open to a re-review? Happy to adjust anything.

packages/app/src/extend.ts
```diff
@@ -78,10 +78,13 @@ export const extendMiddleware = <EngineOptions extends TemplateEngineOptions = T
       if (!acceptsInstance) acceptsInstance = new Accepts(req)
       return acceptsInstance
     }
-    req.accepts = (...types) => getAcceptsInstance().types(types)
-    req.acceptsCharsets = (...charsets) => getAcceptsInstance().charsets(charsets)
-    req.acceptsEncodings = (...encodings) => getAcceptsInstance().encodings(encodings)
-    req.acceptsLanguages = (...languages) => getAcceptsInstance().languages(languages)
+    req.accepts = ((...types: string[]) => getAcceptsInstance().types(types)) as Request['accepts']
+    req.acceptsCharsets = ((...charsets: string[]) =>
+      getAcceptsInstance().charsets(charsets)) as Request['acceptsCharsets']
```
v1rtl, line 83: Does it have to be invoked on every call, or it could be just statically defined and then reused for all Accepts methods?

v1rtl (commented):
Overall good, one tiny question

packages/app/src/extend.ts
claygeo: Good question. Short version: it's already defined once and shared across all four methods, just lazily. Three things going on:

1. **It's already reused across all four methods.** `getAcceptsInstance()` memoizes a single `Accepts` instance in the closure (`if (!acceptsInstance) acceptsInstance = new Accepts(req)`), so `accepts` / `acceptsCharsets` / `acceptsEncodings` / `acceptsLanguages` all share one instance per request. It's never recreated per method or per call.

2. **It can't be hoisted to module scope.** `Accepts` is request-bound: it captures `req.headers` and builds a `Negotiator` over the request. So each request needs its own instance; a module-level singleton would leak one request's headers into the next.

3. **Why it's behind the getter (lazy) rather than an eager `const`.** That's the hot-path optimization from #485. `extendMiddleware` runs on every request, but most requests never call any accepts method, so the lazy getter skips the `new Accepts(req)` + `new Negotiator(req)` allocations entirely for them. When a method *is* called, the getter is just a closure call plus a null check, and the instance is built at most once. (Both `Accepts` and `Negotiator` constructors only store references; no header parsing happens until a method actually runs.)

Worth noting this PR is type-only. The getter predates it (added in #485); I only retyped the four assignment lines so the new overloads narrow correctly (no-arg returns `string[]`, with-arg returns `string | false`). Runtime is unchanged.

If you'd rather drop the getter for readability and use an eager `const acceptsInstance = new Accepts(req)`, I'm happy to do that here. The only tradeoff is it adds those two small allocations back to every request even when content negotiation isn't used, which is why I left #485's lazy form in place. Your call either way.

Verified locally on the branch: `tsc` is clean and the accepts/request/req suites pass (115 tests).

PR #505, opened 2026-07-12
refactor: removed fnc creation in handler fnc to avoid closure creation

exvillager:
### Description
  Refactors request handling in the `App` class so the `loop` and `handle` closures are no longer allocated on every incoming request.
code extracted into private methods (`#loop` and `#handle`) with a small explicit state object.
    
### Benefits
 - Reduces per-request closure allocations (2 fewer function objects created per request).

### Benchmarks
- Tested on my machine (Apple Silicon, Node v24.16.0 `wrk -t4 -c64 -d10s`, hello-world route).
- throughput is the same as master (~83k req/s on both, within run-to-run noise).
- and GC counts under a fixed 500k-request load are essentially identical. 

PS
 - So this PR doesn't make the performance of the application faster than current code.

### DIFF
--- a/packages/app/src/app.ts
+++ b/packages/app/src/app.ts
@@ -306,67 +306,7 @@ export class App<Req extends Request = Request, Res extends Response = Response>
 
     req.baseUrl = ''
 
-    const handle = (middleware: Middleware, pathname: string) => {
-      const { path, handler, regex } = middleware
-
-      let params: URLParams
-
-      try {
-        params = regex ? getURLParams(regex, pathname) : {}
-      } catch (e) {
-        console.error(e)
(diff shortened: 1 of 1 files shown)


exvillager:
@v1rtl  check this PR. 
it improves the application performance and tail latency.

v1rtl:
1) can you squash the biome commit with your code commit
2) do you have benchmarks showing the performance increase?

exvillager:
> 1. can you squash the biome commit with your code commit
 -  Done
> 2. do you have benchmarks showing the performance increase?
- Honestly No , i was wrong in PR making that i claimed it's faster but it's actually same as current code.

Check the PR description i updated about this.

Now it's on you , if you want then you can merge this code or close.

